PR #37, opened 2018-10-08
Ms add samples

martinspetlik:


src/flow_mc.py
```diff
@@ -107,6 +110,8 @@ def __init__(self, mesh_step, level_id=None, config=None, clean=False, parent_fi
         # Element centers of computational mesh.
         self.ele_ids = None
         # Element IDs of computational mesh.
+        self.n_fine_elements = 0
+        self._input_sample = []
```
jbrezmorf: Should be empty dictionary. 
Add comments.

test/01_cond_field/process.py
```diff
@@ -60,14 +63,36 @@ def _extract_result(self, sample_dir):
                             raise Exception("Possitive inflow at outlet region.")
                         total_flux += flux  # flux field
                         found = True
+
```
jbrezmorf, line 57: Make a method for extraction of the profiler time in the parent class FlowMC, it is common to all Simulations based on Flow123d.

src/mlmc/mc_level.py
```diff
@@ -456,7 +473,7 @@ def sample_range(self):
             left = min_sample
         right = min(np.max(fine_sample), quantile_3 + 1.5 * iqr)
 
-        # if left <= 0:
-        #     left = 1e-15
+        if left <= 0:
```
jbrezmorf: ??? This breaks the range for non-positive distributions, like $N(-10,1)$.

src/mlmc/mlmc.py
```diff
@@ -284,6 +282,61 @@ def set_initial_n_samples(self, n_samples=None):
...
+        :param moments_fn: Object providing calculating moments
+        :param pbs: Pbs script generator object
+        :return: None
+        """
+        # Get target levels samples
+        ns_target = self.l_target_samples()
+
+        while True:
+            # Set target number of samples and create simulation samples
+            self.set_target_n_samples(ns_target)
```
jbrezmorf: Seems that this method doesn't exist. It is method of MCLevel

src/mlmc/mlmc.py
```diff
@@ -284,6 +282,61 @@ def set_initial_n_samples(self, n_samples=None):
...
+            self.refill_samples()
+            pbs.execute()
+
+            # Wait until at least half of the target samples are done
+            while any(self.l_finished_samples()[index] < ns_target[index] * 0.5 for index, ns in enumerate(ns_target)):
```
jbrezmorf: Why not use np.any?
while np.any(finished_sample[:] < 0.5 * ns_target[:])

src/mlmc/mlmc.py
```diff
@@ -284,6 +282,61 @@ def set_initial_n_samples(self, n_samples=None):
...
+                pass
+
+            # Estimations - new number of samples from currently collected one
+            n_samples = self.estimate_n_samples_for_target_variance(target_var, moments_fn)
+
+            # If new samples are very close to current target samples then we have our best estimation
+            if all((np.abs(n_samples[index] - ns_tar) / ns_tar) < 0.01
```
jbrezmorf: Is it possible to change this to exact match? Must slightly change adding strategy below.

src/mlmc/mlmc.py
jbrezmorf: Also this is inefficient. Sleep a while and call collect_samples explicitly.

src/mlmc/mlmc.py
```diff
@@ -284,6 +282,61 @@ def set_initial_n_samples(self, n_samples=None):
...
+                   if n_samples[index] >= ns_tar else True for index, ns_tar in enumerate(ns_target)):
+                break
+
+            # New target sample is 10 percent of difference
+            # between current number of target samples and new estimated one
+            new_ns_target = []
+            for n_sample, ns_tar in zip(n_samples, ns_target):
```
jbrezmorf:
```
if np.all( ns_target > 0.5 * n_samples) :
	ns_target = n_samples
```

src/mlmc/mlmc.py
```diff
@@ -284,6 +282,61 @@ def set_initial_n_samples(self, n_samples=None):
...
+    def l_target_samples(self):
+        """
+        Get all levels target number of samples
+        :return: list 
+        """
+        return [level.target_n_samples for level in self.levels]
+
+    def l_finished_samples(self):
+        """
+        Get all levels number of sample values
```
jbrezmorf: This getter has a significant side effect.  Should not be. Make call to collect_sample explicitely in the waiting loop.

src/mlmc/simulation.py
```diff
@@ -34,10 +34,11 @@ def set_coarse_sim(self, coarse_sim=None):
         """
 
     @abstractmethod
-    def simulation_sample(self, tag):
+    def simulation_sample(self, tag, time):
```
jbrezmorf, line 37: Not clear how the generation time and the execution time are distinguished as they are stored as a single time in collected samples.

PR #45, opened 2018-12-26
Ms covariance

jbrezmorf:


src/flow_mc.py
```diff
@@ -279,15 +279,15 @@ def simulation_sample(self, sample_tag, sample_id, start_time=0):
         """
         out_subdir = os.path.join("samples", str(sample_tag))
         sample_dir = os.path.join(self.work_dir, out_subdir)
-        if not os.path.isdir(sample_dir):
-            force_mkdir(sample_dir)
+
+        force_mkdir(sample_dir, True)
```
jbrezmorf, line 283: Is it OK to force overwrite samples?

src/mlmc/estimate.py
```diff
@@ -387,7 +365,6 @@ def best_fit_all(values, range_a, range_b):
                         best_fit_value = fit_value
         return best_fit
 
-
     @staticmethod
     def best_p1_fit(values):
```
jbrezmorf: In fact I made a copy of the code related to construction of density orthogonal moments into simple_distribution.py.
The reason was to test distribution independently of Estimate. So we can remove this stuff from estimate.py.

src/mlmc/estimate.py
```diff
@@ -1263,18 +1226,18 @@ def plot_densities(self, i_sample_mlmc=0):
...
                 continue
             label = "L = {}".format(mc.n_levels)
             distr_plot.add_distribution(mc._distribution, label=label)
 
-        distr_plot.show(save='compare_distributions.pdf')
+        if i_sample_mlmc is not None:
+            mc0_samples = self.mlmc[i_sample_mlmc].levels[0].sample_values[:, 0]
+            distr_plot.add_raw_samples(mc0_samples)
+
+        distr_plot.show(None)#save='compare_distributions.pdf')
```
jbrezmorf: Default should be to save the plot.

src/mlmc/hdf.py
```diff
@@ -170,7 +173,10 @@ class LevelGroup:
                        "result": {'name': 'collected_values', 'default_shape': (0, 2, 1), 'maxshape': (None, 2, None),
                                   'dtype': np.float64},
                        "time": {'name': 'collected_times', 'default_shape': (0, 2, 1), 'maxshape': (None, 2, None),
-                                'dtype': np.float64}}
+                                'dtype': np.float64},
+                       "running_time": {'name': 'running_times', 'default_shape': (0, 2, 1), 'maxshape': (None, 2, None),
+                                'dtype': np.float64}
+                       }
```
jbrezmorf, line 178: We should probably add also some model specific attributes , e.g. mesh step,  mesh size ... However these whould be attributes of level not of the sample.

src/mlmc/mc_level.py
```diff
@@ -285,50 +286,42 @@ def set_coarse_sim(self):
             self.fine_simulation.set_coarse_sim(self.coarse_simulation)
             self.n_ops_estimate = self.fine_simulation.n_ops_estimate()
 
-    def _make_sample_pair(self):
+    def _make_sample_pair(self, sample_pair_id=None):
```
jbrezmorf, line 289: Where is explicit sample ID necessary?

src/mlmc/hdf.py
jbrezmorf: Is HDF5 backward compatible? I.E. can we read  previously generated files? We should be careful about it. Possibly we can collect also times as sort of random values extracted from samples.

src/mlmc/mlmc.py
```diff
@@ -37,6 +34,8 @@ def __init__(self, n_levels, sim_factory, step_range, process_options):
         # Total variance
         self.target_variance = None
 
+        self._estimator = Estimate(self)
```
jbrezmorf, line 37: This leads to sort of circular dependency  we should rather move variance estimation into Estimate, possibly passing the Estimate or its results to sample collecting routines.

src/mlmc/plot.py
```diff
@@ -266,13 +263,14 @@ def add_values(self, values, errors=None, threshold=None, label=""):
...
-            errors = np.flip(errors)
+            values = np.flip(values, axis=0)
+            if errors is not None:
+                errors = np.flip(errors, axis=0)
             threshold = len(values) - 1 - threshold
 
         if self.log_y:
             # plot only positive values
-            i_last_positive = len(values) - np.argmax(np.flip(values) > 0)
+            i_last_positive = len(values) - np.argmax(np.flip(values, axis=0) > 0)
```
jbrezmorf: values should be a vector, we should add an assert possibly

src/pbs.py
```diff
@@ -118,7 +118,7 @@ def execute(self):
         script_content = "\n".join(self.pbs_script)
         pbs_file = os.path.join(self._job_dir, "{:04d}.sh".format(self._job_count))
 
-        pbs_file_pom = os.path.join("/storage/liberec1-tul/home/martin_spetlik",
+        pbs_file_pom = os.path.join("/storage/liberec3-tul/home/martin_spetlik",
```
jbrezmorf: That should not be part of pbs.py but part of its setup

jbrezmorf (commented):
process refactoring is too big change that can not be compared.
We should rather first simplify only one of the two process files and then proceed to the second one.

src/mlmc/mlmc.py
martinspetlik: It will take a while, so I'll create new issue. You can merge it anyway.

PR #49, opened 2019-02-15
Jb ms cov merge

jbrezmorf:


src/mlmc/hdf.py
```diff
@@ -166,26 +172,34 @@ class LevelGroup:
                        "result": {'name': 'collected_values', 'default_shape': (0, 2, 1), 'maxshape': (None, 2, None),
                                   'dtype': np.float64},
                        "time": {'name': 'collected_times', 'default_shape': (0, 2, 1), 'maxshape': (None, 2, None),
-                                'dtype': np.float64}}
+                                'dtype': np.float64},
```
jbrezmorf, line 180: I had to comment out this new attribute in order to process results from 01_data. It seems that all samples are skipped.  A conversion or backward compatibility is necessary.

src/pbs.py
```diff
@@ -118,19 +115,13 @@ def execute(self):
         script_content = "\n".join(self.pbs_script)
         pbs_file = os.path.join(self._job_dir, "{:04d}.sh".format(self._job_count))
 
-        pbs_file_pom = os.path.join("/storage/liberec1-tul/home/martin_spetlik",
-                                    "{:04d}.sh".format(self._job_count))
         self._job_count += 1
-        with open(pbs_file_pom, "w") as file_writer:
+        with open(pbs_file, "w") as file_writer:
```
jbrezmorf, line 119: It seems as workaround for problems with astra share, not necessary anymore. I propose just write into pbs_file directly.

src/mlmc/mlmc.py
```diff
@@ -270,50 +213,51 @@ def set_initial_n_samples(self, n_samples=None):
...
-                                           new_scheduled))
-            # Levels where estimated are greater than scheduled
-            greater_items = np.where(np.greater(n_estimated, n_scheduled))[0]
-
-            # Scheduled samples and wait until at least half of the samples are done
-            self.set_scheduled_and_wait(n_scheduled, greater_items, pbs, sleep)
-
-            # New estimation according to already finished samples
-            n_estimated = np.ceil(np.max(self.estimate_n_samples_for_target_variance(target_var, moments_fn), axis=1))
+    # def target_var_adding_samples(self, target_samples, moments_fn, pbs=None, sleep=20, add_coef=0.1):
```
jbrezmorf: Two methods dependent on Estimate commented. Seems that the first one is not in use. The second one can be easily replaced by setting target n_samples directly. 
More work has to be done in test_mlmc.py

PR #64, opened 2020-02-08
Ms lib design

martinspetlik:


src/mlmc/new_simulation.py
```diff
@@ -1,6 +1,7 @@
+import numpy as np
```
jbrezmorf: why this import

src/mlmc/new_simulation.py
```diff
@@ -1,6 +1,7 @@
...
 import attr
 from abc import ABC, abstractmethod
-from typing import List, Tuple, Union
+from typing import List, Tuple, Union, Any
```
jbrezmorf: Any not used?

src/mlmc/new_simulation.py
```diff
@@ -11,6 +12,7 @@ class QuantitySpec:
     shape: Tuple[int, int]
     times: List[float]
     locations: Union[List[str], List[Tuple[float, float, float]]]
+    used_attributes: List = ["name", "unit", "shape", "times", "locations"]
```
jbrezmorf: ??

src/mlmc/sampling_pool.py
```diff
@@ -51,39 +53,133 @@ def compute_seed(sample_id):
...
+    def _change_to_sample_directory(self, path: str):
+        """
+        Create sample directory and change working directory
+        :param path: str
+        :return: None
+        """
+        sample_dir = os.path.join(self._work_dir, path)
+        if not os.path.isdir(sample_dir):
+            os.makedirs(sample_dir, mode=0o775, exist_ok=True)
+        os.chdir(sample_dir)
```
jbrezmorf, line 82: Call chdir just before call of the simulation (i.e. in calculate_sample) and then 
change dir back. Consider a context manager:
```
class change_cwd:
    """
    Context manager that change CWD, to given relative or absolute path.
    """
    def __init__(self, path: str):
        self.path = path
        self.orig_cwd = ""

    def __enter__(self):
        if self.path:
            self.orig_cwd = os.getcwd()
            os.chdir(self.path)

    def __exit__(self, exc_type, exc_value, traceback):
        if self.orig_cwd:
            os.chdir(self.orig_cwd)
```

jbrezmorf: Also consider rename of the method.

jbrezmorf:
Opravy mimo chdir, a pak merge do master.

PR #65, opened 2020-02-12
Ms new design pbs

martinspetlik:
- sampling pool: _scheduled_samples_(level_sim, sample_id)
			- serialize level_sim
			- copy files to sample_directory (which is created at that moment)
                        - compute seed
                        - append (level_sim.level_id, sample_id, seed) to _scheduled list
                        - add job weight, increment number of samples in job
                           _execute_
			- it is call when job weight or number of samples per job is exceed
			- create files: {jobs}/XYZ_job.sh, {jobs}/XYZ_scheduled.yaml, {jobs}/XYZ_pbs_id
                        - format pbs script
                        - run pbs file

- pbs_process - run in pbs file:
               get_level_config()
			 - deserialization of level_sim
                         - read scheduled
	       calculate_samples()
			 - level_sim.calculate()
                         - save success, failed, times for success
               write_results_to_file(success, failed, times)


src/mlmc/new_simulation.py
```diff
@@ -2,7 +2,7 @@
 import attr
 from abc import ABC, abstractmethod
 from typing import List, Tuple, Union, Any
-from mlmc.level_simulation import LevelSimulation
+from level_simulation import LevelSimulation
```
jbrezmorf: mlmc. missing

src/mlmc/sampling_pool_pbs.py
```diff
@@ -0,0 +1,459 @@
...
+    TIME = "{}_times.yaml"
+    PBS_ID = "{}_"
+    JOB = "{}_job.sh"
+    LEVEL_SIM_CONFIG = "level_simulation_config"
+    LEVELS_CONFIG = "levels_config.txt"
+    STRUCTURE = "structure.json"
+
+    def __init__(self, job_weight=200000, job_count=0, work_dir=None):
+        """
+        :param job_weight: Number of simulation elements per job script
```
jbrezmorf: I see no usage for the 'job_weight' parameter as the weight is level dependent.
For the other two arguments, I don't think they are necessary. 
Provide better documentation of their purpose if you think they are.

src/mlmc/sampling_pool_pbs.py
```diff
@@ -0,0 +1,459 @@
...
+        :param job_count: Number of created jobs
+        """
+        self._work_dir = work_dir
+        # Weight of the single PBS script (putting more small jobs into single PBS job).
```
jbrezmorf: Attribute comments should be under related assignments.

src/mlmc/sampling_pool_pbs.py
```diff
@@ -0,0 +1,459 @@
...
+        self._number_of_realizations = 0
+        # Lines to put at the beginning of the PBS script.
+        self.pbs_script = None
+        self._pbs_config = None
+        self._pbs_header_template = None
+
+        self._scheduled = []
+        
+        # Level directories
+        self._level_dir = {}
```
jbrezmorf: Why this is a dictionary?

src/mlmc/sampling_pool_pbs.py
```diff
@@ -0,0 +1,459 @@
...
+    OUTPUT_DIR = "output"
+    LEVEL_DIR = "level_{}"
+    JOBS_DIR = "jobs"
+    SCHEDULED = "{}_scheduled.yaml"
+    SUCCESSFUL_RESULTS = "{}_successful_results.yaml"
+    FAILED_RESULTS = "{}_failed_results.yaml"
+    TIME = "{}_times.yaml"
+    PBS_ID = "{}_"
+    JOB = "{}_job.sh"
+    LEVEL_SIM_CONFIG = "level_simulation_config"
```
jbrezmorf: Can we use some extension here?

src/mlmc/sampling_pool_pbs.py
```diff
@@ -0,0 +1,459 @@
...
+class SamplingPoolPBS(SamplingPool):
+
+    OUTPUT_DIR = "output"
+    LEVEL_DIR = "level_{}"
+    JOBS_DIR = "jobs"
+    SCHEDULED = "{}_scheduled.yaml"
+    SUCCESSFUL_RESULTS = "{}_successful_results.yaml"
+    FAILED_RESULTS = "{}_failed_results.yaml"
+    TIME = "{}_times.yaml"
+    PBS_ID = "{}_"
```
jbrezmorf: What is the purpose? Usage in write_pbs_id is confusing or wrong?

src/mlmc/sampling_pool_pbs.py
```diff
@@ -0,0 +1,459 @@
...
+        self._create_output_dir()
+        self._create_job_dir()
+
+    def _create_output_dir(self):
+        """
+        Create output dir in working directory, remove existing output dir
+        :return: None
+        """
+
+        if self._work_dir is None:
```
jbrezmorf: Rather check in in init and do not provide the default value.

src/mlmc/sampling_pool_pbs.py
```diff
@@ -0,0 +1,459 @@
...
+            raise NotADirectoryError("Working directory must be pass to Sampler init")
+
+        if self._output_dir is None:
```
jbrezmorf: Why the check if it is created in the __init__.

src/mlmc/sampling_pool_pbs.py
```diff
@@ -0,0 +1,459 @@
...
+
+    @property
+    def jobs_dir(self):
+        if self._jobs_dir is None:
+            self._create_output_dir()
+            self._create_job_dir()
+        return self._jobs_dir
+
+    @property
+    def pbs_job_file(self):
```
jbrezmorf: I don't understand the purpose of these properties. 
Anyway, the exception is appropriate only when openning actual file.

src/mlmc/sampling_pool_pbs.py
```diff
@@ -0,0 +1,459 @@
...
+        if self._files_structure is None:
+            raise FileNotFoundError
+        return self._files_structure
+
+    def _create_job_dir(self):
+        if self._work_dir is None:
+            raise NotADirectoryError("Working directory must be pass to Sampler init")
+
+        if self._jobs_dir is None:
+            self._jobs_dir = os.path.join(self._output_dir, SamplingPoolPBS.JOBS_DIR)
```
jbrezmorf: Can be set in __init__ or need not to be checked?

src/mlmc/sampling_pool_pbs.py
```diff
@@ -0,0 +1,459 @@
...
+    def read_results(self, job_id):
+        with open(os.path.join(self._jobs_dir, SamplingPoolPBS.RESULTS.format(job_id))) as reader:
+            reader.readlines()
+
+    def _save_structure(self):
+        """
+        Save structure of files needed for pbs process run
+        :return: None
+        """
+        files_structure = {"scheduled": os.path.join(self._jobs_dir, SamplingPoolPBS.SCHEDULED),
```
jbrezmorf: Only time has to be saved, others are constant?

src/mlmc/sampling_pool_pbs.py
```diff
@@ -0,0 +1,459 @@
...
+import re
+from typing import List
+import pickle
+import json
+import glob
+from level_simulation import LevelSimulation
+from sampling_pool import SamplingPool
+
+
+class SamplingPoolPBS(SamplingPool):
```
jbrezmorf, line 13: We should introduce a class for a single Job.

src/mlmc/sampling_pool_pbs.py
```diff
@@ -0,0 +1,459 @@
...
+                                     '#PBS -S /bin/bash',
+                                     '#PBS -l select={n_nodes}:ncpus={n_cores}:mem={mem}{select_flags}',
+                                     '#PBS -l walltime=1:00:00',
+                                     '#PBS -q {queue}',
+                                     '#PBS -N MLMC_sim',
+                                     '#PBS -j oe',
+                                     '#PBS -o {pbs_output_dir}/{job_name}.OU',
+                                     '#PBS -e {pbs_output_dir}/{job_name}.ER',
+                                     '']
+        if flow_3:
```
jbrezmorf: This should be part of the simulation.

src/mlmc/sampling_pool_pbs.py
```diff
@@ -0,0 +1,459 @@
...
+        """
+        if self._output_dir is not None:
+            if not os.path.isdir(os.path.join(self._output_dir, SamplingPoolPBS.LEVEL_DIR.format(level_id))):
+                level_dir = os.path.join(self._output_dir, SamplingPoolPBS.LEVEL_DIR.format(level_id))
+                os.makedirs(level_dir, mode=0o775, exist_ok=True)
+                self._level_dir[level_id] = level_dir
+        else:
+            raise NotADirectoryError("Create output directory at the first place")
+
+    def change_to_sample_directory(self, path: str, level_id=None):
```
jbrezmorf: That should be part of pbs_process.

src/mlmc/sampling_pool_pbs.py
```diff
@@ -0,0 +1,459 @@
...
+        :param level_id: str
+        :param sample_id: str
+        :return: None
+        """
+        sample_dir = os.path.join(self._level_dir[level_id], sample_id)
+        if not os.path.isdir(sample_dir):
+            os.makedirs(sample_dir, mode=0o775, exist_ok=True)
+        os.chdir(sample_dir)
+
+    def _copy_sim_files(self, files: List[str]):
```
jbrezmorf: Something similar above.

src/mlmc/sampling_pool_pbs.py
```diff
@@ -0,0 +1,459 @@
...
+
+            # Get number of running samples
+            if pbs_id in unfinished_pbs_jobs:
+                with open(os.path.join(self._jobs_dir, SamplingPoolPBS.SCHEDULED.format(job_id))) as file:
+                    lines = yaml.load(file, yaml.Loader)
+                    n_running += len(lines)
+                continue
+
+            # Get sample results
+            for level_id, path in self._level_dir.items():
```
jbrezmorf: Possibly a single job/a single pbs_process could write the results into a single directory or even into a single file (with internal structure per level). That simplifies the file structure (no level dirs) and can speed up data IO. 
We can possibly write a sort of mark at the end of the result file in order to check that it is correctly synchronized through the NFS.

src/mlmc/pbs_process.py
```diff
@@ -0,0 +1,158 @@
...
+from pathlib import Path
+
+
+src_path = os.path.dirname(os.path.abspath(__file__))
+sys.path.append(os.path.join(src_path, '..', '..', 'src'))
+sys.path.append(os.path.join(src_path, '..', '..', 'test'))
+sys.path.append(os.path.join(src_path))
+
+
+class PbsProcess:
```
jbrezmorf, line 11: Possibly we can view part of the proposed 'Job' class (see below) and the core of PbsProcess as a single object. Created in SamplingPoolPBS by adding the samples, then serialized, passed to qsub, and deserialized and executed in pbs_process.py. Parsing the arguments can be a separte function and we can have a class method as a factory. None initialization is boring.

jbrezmorf (commented):
The recipe is fine, I have some questions and suggestions to the implementation details.

src/mlmc/new_simulation.py
martinspetlik: I use relative imports because of Pickle

src/mlmc/sampling_pool_pbs.py
martinspetlik: SamplingPool know nothing about number of levels.

src/mlmc/sampling_pool_pbs.py
martinspetlik: job_weight is actually the maximum sum of task sizes of all samples in the job, so if the sum exceed the maximum then the job is executed.

job_count would be useful for regenerating failed samples while some old samples are still running, so it should provide unique job id.

without work_dir it cannot work in current form

I will make doc better.

src/mlmc/sampling_pool_pbs.py
martinspetlik: write_pbs_id  creates empty file with the following name: jobID_pbsID. It matches jobId with pbsId.

src/mlmc/sampling_pool_pbs.py
martinspetlik: What type of extension do you mean?

src/mlmc/sampling_pool_pbs.py
martinspetlik: It is files structure for pbs_process, which should know location of essential files.

src/mlmc/sampling_pool_pbs.py
jbrezmorf: Add list of modules to load into PBSPool configuration. 
Add also path to the python to use (with default 'python3').

src/mlmc/pbs_process.py
```diff
@@ -0,0 +1,158 @@
+import os
+import sys
+import yaml
+import time
+import pickle
+import json
+from pathlib import Path
+
+
+src_path = os.path.dirname(os.path.abspath(__file__))
```
jbrezmorf: These sys.path hacks should not appear in production code.
Use virtual environment, install mlmc in it, and use it consistently in PBS jobs.

src/mlmc/sampling_pool_pbs.py
```diff
@@ -0,0 +1,311 @@
...
+
+
+class SamplingPoolPBS(SamplingPool):
+
+    OUTPUT_DIR = "output"
+    JOBS_DIR = "jobs"
+    LEVEL_SIM_CONFIG = "level_{}_simulation_config"  # Serialized level simulation
+    JOB = "{}_job.sh"  # Pbs process file
+
+    def __init__(self, work_dir, job_weight=200000, job_count=0):
```
jbrezmorf, line 20: This constructor is called by the user, who has no idea how to choose job_weight. We should limit the job wall time, which is already part of the PBS configuration. Then the weight of individual samples can be given by n_ops, ie.e their time estimate. The system should be robust against killed jobs, so never mind if the actual job wall time will be a little larger then estimated.

jbrezmorf: job_count can not be provided at the time of construction can be possibly set through the calls of have_permanent_sample.

test/sampler_test_hdf.py
```diff
@@ -7,9 +7,10 @@
 src_path = os.path.dirname(os.path.abspath(__file__))
```
jbrezmorf: Make this a regular test.

test/synth_simulation_workspace.py
```diff
@@ -125,7 +125,23 @@ def calculate(config, seed):
         if np.isnan(fine_result) or np.isnan(coarse_result):
             raise Exception("result is nan")
 
-        return fine_result, coarse_result
+        quantity_format = SynthSimulationWorkspace.result_format()
```
jbrezmorf: I propose to have both synth_simulation and synth_simulation_wrokspace in a single file, and reuse their common code. Remove or rename synth_simulation_test as that is expected to be a test source file by the pytest.

test/test_sampler_pbs.py
```diff
@@ -0,0 +1,107 @@
+import os
```
jbrezmorf, line 1: Introduce a [pytest marker](http://doc.pytest.org/en/latest/example/markers.html) 'pbs' that depends on the PBS system. Make a script in the test directory to run all pbs tests. Skip pbs tests on CircleCI.

test/test_sampler_sim_workspace.py
```diff
@@ -10,16 +10,18 @@
 from synth_simulation_workspace import SynthSimulationWorkspace
```
jbrezmorf: We should make tests composable and test over a matrix:
(several different simulations) x (different pools) x (different storage)

jbrezmorf (commented):
Fix or discuss remaining problems. Changes in the tests can be done even after the merge, just make an issue for it in such a case. Use issues in general to keep track of the remaining problems.

src/mlmc/sampling_pool_pbs.py
martinspetlik: job_count is removed from constructor in child branch (MS_new_design_failed), so I would make these changes there

PR #70, opened 2020-03-15
Ms new design failed

martinspetlik:


src/mlmc/sample_storage.py
```diff
@@ -72,6 +72,12 @@ def get_n_ops(self):
         :return: List[float]
         """
 
+    def unfinished_ids(self):
+        """
+        Get unfinished sample's ids
+        :return: list
+        """
+
```
jbrezmorf, line 81: missing pass ?

src/mlmc/sampling_pool_pbs.py
```diff
@@ -16,21 +16,20 @@ class SamplingPoolPBS(SamplingPool):
     JOBS_DIR = "jobs"
     LEVEL_SIM_CONFIG = "level_{}_simulation_config"  # Serialized level simulation
     JOB = "{}_job.sh"  # Pbs process file
+    JOBS_COUNT = "jobs_count.txt" # Contains current number of jobs which is also job unique identifier
 
-    def __init__(self, work_dir, job_weight=200000, job_count=0):
+    def __init__(self, work_dir, job_weight=200000, force=False):
```
jbrezmorf: rename 'force' to 'clean'

test/process_test.py
```diff
@@ -0,0 +1,271 @@
...
+        """
+        import argparse
+        parser = argparse.ArgumentParser()
+
+        parser.add_argument('command', choices=['run', 'collect', 'resurrect'], help='run - create new execution,'
+                                                                                     'collect - keep collected'
+                                                                                     'resurrect - resurrect failed samples')
+        parser.add_argument('work_dir', help='Work directory')
+        parser.add_argument("-f", "--force", default=False, action='store_true', help="Clean before run")
+        parser.add_argument("-c", "--clean", default=False, action='store_true', help="Clean auxiliary structures")
```
jbrezmorf: - Exact difference between 'clean' and 'force'?
- rename 'force' to 'clean'?
- which auxiliary structures?, sample dictionaries?
- possibly clean by default, only keep on special option, e.g 'keep-tmp' (keep temporaries, useful for debugging)

jbrezmorf: ProcessPBS can be a later generalize to a general Process class for process script implementation.

test/process_test.py
```diff
@@ -0,0 +1,271 @@
...
+            self.append = True
+            self.force = False
+            self.run()
+        elif args.command == 'resurrect':
+            print("---RESURRECT")
+            self.force = False
+            self.append = True
+            self.run(resurrect=True)
+
+    def get_arguments(self, arguments):
```
jbrezmorf: Move common parts to a generic class.

test/process_test.py
```diff
@@ -0,0 +1,271 @@
...
+                                                                                     'resurrect - resurrect failed samples')
+        parser.add_argument('work_dir', help='Work directory')
+        parser.add_argument("-f", "--force", default=False, action='store_true', help="Clean before run")
+        parser.add_argument("-c", "--clean", default=False, action='store_true', help="Clean auxiliary structures")
+
+        args = parser.parse_args(arguments)
+
+        return args
+
+    def run(self, resurrect=False):
```
jbrezmorf: Where we set number of levels?

src/mlmc/hdf5.py
```diff
@@ -369,14 +369,35 @@ def get_finished_ids(self):
...
     def get_failed_ids(self):
         """
         Failed samples ids
-        :return: set() of failed sample ids
+        :return: list of failed sample ids
         """
         with h5py.File(self.file_name, 'r') as hdf_file:
-            # Return NumPy array otherwise return set()
-            return set(hdf_file[self.level_group_path][self.failed_dset])
+            failed_ids = [sample[0].decode() for sample in hdf_file[self.level_group_path][self.failed_dset][()]]
```
jbrezmorf, line 386: So HDF5 store only bytearrays?

src/mlmc/sampler.py
```diff
@@ -241,3 +262,21 @@ def l_scheduled_samples(self):
         :return: list
         """
         return self._n_target_samples
+
+    def resurrect_failed(self):
+        """
+        Resurrect failed samples
```
jbrezmorf, line 268: Ressurect is possibly not the best name as the new samples are scheduled with the same IDs. I can not recall: Why we need to use the same IDs? Using new IDs we can eliminate modification of HDF data, which is preferred.

src/mlmc/sampling_pool_pbs.py
```diff
@@ -16,21 +16,20 @@ class SamplingPoolPBS(SamplingPool):
     JOBS_DIR = "jobs"
     LEVEL_SIM_CONFIG = "level_{}_simulation_config"  # Serialized level simulation
     JOB = "{}_job.sh"  # Pbs process file
+    JOBS_COUNT = "jobs_count.txt" # Contains current number of jobs which is also job unique identifier
```
jbrezmorf, line 55: If we need this, can we store with some other information, not introducing a new file for that?
But possibly we can get this only by listing the 'jobs' directory and getting all "*_job.sh" job scripts.

src/mlmc/sampling_pool_pbs.py
```diff
@@ -68,6 +74,18 @@ def _create_job_dir(self):
         self._jobs_dir = os.path.join(self._output_dir, SamplingPoolPBS.JOBS_DIR)
         os.makedirs(self._jobs_dir, mode=0o775, exist_ok=True)
 
+    def _get_job_count(self):
```
jbrezmorf, line 113: Use jobs listing, e.g. using the `glob` module.

src/mlmc/sampling_pool_pbs.py
```diff
@@ -109,6 +127,7 @@ def pbs_common_setting(self, **kwargs):
                                      '']
 
         self._pbs_header_template.extend(('cd {work_dir}',
+                                          'module load python36-modules-gcc',
```
jbrezmorf: This whole is  Charon specific. We should make it generic and introduce a sort of injection from the PBS configuration.

src/mlmc/sampling_pool_pbs.py
```diff
@@ -286,23 +308,72 @@ def get_result_files(self, finished_pbs_jobs, unfinished_pbs_jobs):
...
+        Collect samples which had finished after main process crashed, append them to new collected samples
+        :param successful_results: dict
+        :param failed_results: dict
+        :param times: dict
+        :return: all input dictionaries
         """
+        already_collected = []
 
+        for sample_id in self._unfinished_sample_ids:
+            if sample_id in already_collected:
```
jbrezmorf, line 379: Why? Can the samples appear multiple times in the unfinished list?
If necessary use a set instead of the list to prevent a square complexity.

src/mlmc/sampling_pool_pbs.py
```diff
@@ -286,23 +308,72 @@ def get_result_files(self, finished_pbs_jobs, unfinished_pbs_jobs):
...
+
+            sample_dir = os.path.join(self._output_dir, sample_id)
+            file_name = os.path.join(sample_dir, PbsProcess.PERMANENT_SAMPLE.format("*"))
+            file = glob.glob(file_name)[0]
+            job_id = re.findall(r'._(\d+)', file)[0]
+
+            successful, failed, time = PbsProcess.read_results(job_id, self._jobs_dir)
+
+            # Split results to levels
+            for level_id, results in successful.items():
```
jbrezmorf, line 388: This should be already done somewhere, can not we reuse that code?

src/mlmc/synth_simulation.py
```diff
@@ -288,18 +288,4 @@ def _read_config():
         return config
```
jbrezmorf, line 289: Should we have SyntSimulationn as part of the library or rather move it into tests?

src/mlmc/sample_storage_hdf.py
```diff
@@ -151,6 +156,35 @@ def n_finished(self):
...
+
+    def failed_samples(self):
+        """
+        Dictionary of failed samples
+        :return: dict
+        """
+        failed_samples = {}
+
+        for level in self._level_groups:
+            print("level.get_failed_ids() ", level.get_failed_ids())
```
jbrezmorf, line 179: Remove debugging prints. Should we introduce a logging?

src/mlmc/sample_storage_hdf.py
```diff
@@ -151,6 +156,35 @@ def n_finished(self):
...
         return n_finished
 
+    def unfinished_ids(self):
+        """
+        List of unfinished ids
+        :return: list
+        """
+        unfinished = []
+
+        for level in self._level_groups:
```
jbrezmorf, line 166: Seems that `SampleStorageHDF` is aware of the HDF structure, so I ask where is the border between SampleSTORAGEHDF and the hdf5? There are historic reasons, but could we simplify things putting it into a single class?

src/mlmc/sample_storage_hdf.py
```diff
@@ -151,6 +156,35 @@ def n_finished(self):
...
+    def failed_samples(self):
+        """
+        Dictionary of failed samples
+        :return: dict
+        """
+        failed_samples = {}
+
+        for level in self._level_groups:
+            print("level.get_failed_ids() ", level.get_failed_ids())
+            failed_samples[str(level.level_id)] = list(level.get_failed_ids())
```
jbrezmorf, line 180: Why is this dictionary indexed by the level ID as a string?

src/mlmc/pbs_process.py
```diff
@@ -126,16 +128,17 @@ def calculate_samples(self):
...
 
         # List of Tuple[level id, sample id, random seed]
         level_id_sample_id_seed = self._get_level_id_sample_id_seed()
 
         failed = {}
         success = {}
         current_level = 0
+        current_samples = []
         start_time = time.time()
         times = {0: [0, 0]}
```
jbrezmorf: Comment the attributes.

src/mlmc/pbs_process.py
```diff
@@ -210,6 +231,14 @@ def _write_end_mark(self, path):
                 yaml.dump(data, f)
 
 
+    def _current_samples_to_permanent(self, current_samples):
+        for sample_id in current_samples:
+            self._change_to_sample_directory(sample_id)
+
+            file_name = os.path.join(os.getcwd(), PbsProcess.PERMANENT_SAMPLE.format(self._job_id))
```
jbrezmorf: Marking the permanent samples by the file can be quite slow, is it necessary?

src/mlmc/sampling_pool_pbs.py
```diff
@@ -286,23 +308,72 @@ def get_result_files(self, finished_pbs_jobs, unfinished_pbs_jobs):
...
+        :return: all input dictionaries
         """
+        already_collected = []
 
+        for sample_id in self._unfinished_sample_ids:
+            if sample_id in already_collected:
+                continue
+
+            sample_dir = os.path.join(self._output_dir, sample_id)
+            file_name = os.path.join(sample_dir, PbsProcess.PERMANENT_SAMPLE.format("*"))
```
jbrezmorf: No need to mark permanent samples, need to review POOL API but the scheduler can in principle deal with permanent sample ids that were already in the storage.

jbrezmorf (commented):
Several things to discus, few to improve.

src/mlmc/hdf5.py
jbrezmorf: Yes.

src/mlmc/pbs_process.py
jbrezmorf: 1. It is done at the job level, so in paralel and not critical.
2. Resolve issue with job files appending, using plain list instead of the dictionary for level ids.
3. Keep premanent sample file but understand it as an implementation of a permanent map form a sample_id to the job_id. Make separate methods for writing to the map and reading from it.

src/mlmc/sample_storage.py
jbrezmorf: OK. Doc string is enough.

src/mlmc/sample_storage_hdf.py
jbrezmorf: Make an issue for this:
- either merge
- o abstract SampleStorageHDF to SampleStoragePermanent and make a general API of HDF5 independent on particualr datbase format.

src/mlmc/sampler.py
jbrezmorf: The reason is that sample ID determines (through hash) the seed for the random numbers of the sample.
Make an issue:
- either separate sample_id and seed
- or create a new HDF5 during ressurection reusing results of the original file

src/mlmc/sampling_pool_pbs.py
jbrezmorf: Should we have a clean flag also for multiprocess pool?

src/mlmc/sampling_pool_pbs.py
jbrezmorf: Assume that MLMC package is installed, call pbs_process as:

    python -m mlmc.pbs_process > ...

Let 'user' to choose how this is done.  We can do it in the main process script.

src/mlmc/sampling_pool_pbs.py
jbrezmorf: Just use the set.

src/mlmc/sampling_pool_pbs.py
jbrezmorf: See above.

test/process_test.py
```diff
@@ -0,0 +1,271 @@
...
+        parser = argparse.ArgumentParser()
+
+        parser.add_argument('command', choices=['run', 'collect', 'resurrect'], help='run - create new execution,'
+                                                                                     'collect - keep collected'
+                                                                                     'resurrect - resurrect failed samples')
+        parser.add_argument('work_dir', help='Work directory')
+        parser.add_argument("-f", "--force", default=False, action='store_true', help="Clean before run")
+        parser.add_argument("-c", "--clean", default=False, action='store_true', help="Clean auxiliary structures")
+
+        args = parser.parse_args(arguments)
```
jbrezmorf: Rename PBSProcess to PBSJob to avoid name conflict.

src/mlmc/sampling_pool_pbs.py
martinspetlik: I don't think we need a clean flag. Because in case of multiprocess pool there are no auxiliary files which are process dependent, all is mainly related to simulation.

src/mlmc/synth_simulation.py
martinspetlik: It should be here, I think there was some error around pickle if it wasn't in mlmc module.

PR #97, opened 2020-08-11
Ms enh

jbrezmorf:


martinspetlik:
Improvements:

- `mlmc.tool.pbs_job` uses `SamplingPool.calculate_sample` method and other methods from SamplingPool base class

- `MLMC_WORKDIR` variable is introduced (for usage see `mlmc.sampling_pool_pbs.SamplingPoolPBS.pbs_common_setting()` method)
	
- new PBS config parameters: `pbs_name`, `walltime`, `optional_pbs_requests` are introduced, for usage see `test/01_cond_field/process_simple.create_sampling_pool()`

- config parameter is removed from `mlmc.sim.Simulation` init

- failed results include exception traceback, according to PE_wgc

- debug mode, if `--debug` param is set, then SamplingPool keeps sample directories
	
- `job_weight` parameter removed from `mlmc.sampling_pool_pbs.SamplingPoolPBS` init, PBS job size is driven by `task_size`, `job_weight` is internally set to 1, so if sum of sample task sizes is above 1, the job is scheduled

- SampleStorageHDF souldn't fail on `sample_pairs() `for option 'append=True' 

-  `mlmc.sim.Simulation.result_format()` is no longer static method, result_format is part of `mlmc.level_simulation.LevelSimulation` and it is set in Sampler

- result format is checked in `mlmc.sampling_pool.SampingPool.calculate()`

- output directory is created for all sampling pools (if needs workspace)
		

Not fixed issue:
   _missing failed results yaml file when running locally_ - that is expected behavior

PR #98, opened 2020-08-25
Ms quantity distributions

martinspetlik:
MS_quantity branch contains code for approximation of the distribution function using the maximum entropy method and also using BSpline and other stuff which were used for the Diploma Thesis.

Quantity is supported.


src/mlmc/quantity_concept.py
```diff
@@ -62,28 +189,39 @@ def __init__(self, input_quantities, operation = None):
...
+            quantity.set_range(start, size)
+
+    def size(self) -> int:
+        return self.qtype.size()
+
+    def get_cache_key(self, level_id, i_chunk):
+        return str(hash((level_id, i_chunk, id(self), *[id(q) for q in self._input_quantities])))
 
-    def _samples(self, i_chunk):
+    @clearable_cache(custom_key_maker=get_cache_key)
```
jbrezmorf: ? why clearable
? can we use max_size=1
? simplification of get_cache_key:
  - just return tuple
  - no need to hash also input quantities as these are unique for self anyway

jbrezmorf (commented):
Push fresh commits

src/mlmc/moments.py
```diff
@@ -77,6 +77,9 @@ def __call__(self, value):
     def eval(self, i, value):
         return self._eval_all(value, i+1)[:, -1]
 
+    def eval_fine_coarse(self, i, value):
```
jbrezmorf: K cemu se toto pouziva? Je to neefektivni pocitat vsechny momenty az do i a pak vybrat jen ten posledni.

src/mlmc/sample_storage.py
```diff
@@ -187,14 +200,30 @@ def load_scheduled_samples(self):
     def sample_pairs(self):
         """
         Sample results split to numpy arrays
+        :param i_chunk: identifier of chunk
```
jbrezmorf: Ma `sample_pairs` jeste smysl? Zejmena kdyz neni podporovano `i_chunk`?

src/mlmc/sample_storage_hdf.py
```diff
@@ -145,16 +145,28 @@ def sample_pairs(self):
...
         return levels_results
 
+    def sample_pairs_level(self, level_id, i_chunk=0):
+        """
+        Get result for particular level and chunk
+        :param level_id: int, level id
+        :param i_chunk: int, chunk id
+        :return:
+        """
+        if i_chunk != 0:
```
jbrezmorf: To je nejake divne, nema tu byt:
`if i_chunk==0:`?
Podobne pro memory storage. 
Mas nejaky test kde se pocita s vice chunkama?

src/mlmc/quantity_estimate.py
```diff
@@ -223,30 +223,27 @@ def estimate_diff_mean(self, moments_fn, level_result, zero_level=False):
...
-    #         mom_sum = mom_fine + mom_coarse
-    #         cov = 0.5 * (np.matmul(mom_diff.T, mom_sum) + np.matmul(mom_sum.T, mom_diff)) / self.n_samples
-    #     else:
-    #         # Direct formula
-    #         cov_fine = np.matmul(mom_fine.T,   mom_fine)
-    #         cov_coarse = np.matmul(mom_coarse.T, mom_coarse)
-    #         cov = (cov_fine - cov_coarse) / self.n_samples
-    #
-    #     return cov
+    def estimate_covariance(self, moments_fn, stable=False):
```
jbrezmorf: We should do the cov matrix estimate through quantities and the estimate_mean algorithm.
The stable summation algorithm tries to reduce arithmetic errors. That can also be an issue of the general estimation algorithm. See, e.g. https://en.wikipedia.org/wiki/Kahan_summation_algorithm

That can have some effect if we try to obtain very precise estimates using millions of samples.

src/mlmc/sim/synth_simulation.py
```diff
@@ -116,7 +116,10 @@ def calculate(config, seed):
         for result in [fine_result, coarse_result]:
```
jbrezmorf: General not to `synth_simulation.py` currently this is just a test mock class and should be in test directory possibly as a simple tutorial. But we can possibly implement a SyntheticSimulation based on given exact distribution.

src/mlmc/tool/hdf5.py
```diff
@@ -197,6 +196,8 @@ class LevelGroup:
     COLLECTED_ATTRS = {"sample_id": {'name': 'collected_ids', 'default_shape': (0,), 'maxshape': (None,),
                                      'dtype': SCHEDULED_DTYPE}}
 
+    chunk_size = 512000000  # bytes in decimal
```
jbrezmorf: The chunk_size should rather be attribute of the storage class as it reflects size of the computational space in the memory which is independent of the actual storage file format.

test/development_tests.py
```diff
@@ -21,24 +21,24 @@ def sampler_hdf_test():
     failed_fraction = 0.1
 
     work_dir = os.path.join(os.path.dirname(os.path.realpath(__file__)), '_test_tmp')
-    # if os.path.exists(work_dir):
-    #     shutil.rmtree(work_dir)
-    # os.makedirs(work_dir)
+    if os.path.exists(work_dir):
```
jbrezmorf: Why this file name? We should organize tests according to the files of the library.

src/mlmc/quantity_concept.py
```diff
@@ -1,109 +1,313 @@
...
+    :param method: numpy method
+    :return: inner method
+    """
+    def _method(*args):
+        """
+        To process input quantities to perform numpy method
+        :param args: args
+        :return: Quantity
+        """
+        if all(isinstance(quantity, Quantity) for quantity in args):
```
jbrezmorf: Should first convert constant arguments (flat, bool, Array, ..) to Quantity.

jbrezmorf: We should have a policy for error reporting. Good error reporting is extremely helpful. 
Separate issue #101.

src/mlmc/quantity_concept.py
```diff
@@ -1,109 +1,313 @@
...
+                "All quantities must have same base type either ScalarType or BoolType"
+
+            return Quantity(quantity_type=args[0].qtype, input_quantities=custom_copy(list(args)), operation=method)
+        else:
+            raise Exception("All parameters must be quantities")
+    return _method
+
+
+# Numpy methods assigned to this module, this methods work with quantities
+for method in [np.sin, np.cos, np.add, np.maximum, np.logical_and, np.logical_or]:
```
jbrezmorf: Seems there is [a way](https://numpy.org/doc/stable/reference/arrays.classes.html#arrays-classes) to do this for all [`numpy.unfunc`](https://numpy.org/doc/stable/reference/ufuncs.html) at once. Can be postponed to separate issue, but definitely we need support for more functions.

src/mlmc/quantity_concept.py
```diff
@@ -1,109 +1,313 @@
...
+    dict_types = []
+    for q_spec in q_specs:
+        scalar_type = ScalarType(float)
+        array_type = ArrayType(q_spec.shape, scalar_type)
+        field_type = FieldType([(loc, array_type) for loc in q_spec.locations])
+        ts_type = TimeSeriesType(q_spec.times, field_type)
+        dict_types.append((q_spec.name, ts_type))
+    dict_type = DictType(dict_types)
+
+    return Quantity(quantity_type=dict_type, input_quantities=[QuantityStorage(storage, dict_type)])
```
jbrezmorf: Why not just `return QuantityStorage(storage, dict_type)`?

src/mlmc/quantity_concept.py
```diff
@@ -1,109 +1,313 @@
...
     """
-    Concept of the MLMC mean estimator.
-    TODO:
-    - test
-    - calculate variance estimates as well
-    :param quantity:
-    :return:
+    MLMC mean estimator
+    :param quantity: Quantity
+    :return: QuantityMean which holds both mean and variance
```
jbrezmorf: This is the key function. Try to provide a better description. 
In particular we should be more specific about resulting mean and variance.
E.g. :
```
MLMC mean estimator. The MLMC method is used to compute the mean estimate
to the Quantity dependent on the collected samples. The squared error of the estimate (the estimator variance) is estimated using the central limit theorem.
```

src/mlmc/quantity_concept.py
```diff
@@ -1,109 +1,313 @@
...
     """
-    quantity_vec_size = quantity._size()
-    n_levels = 1
+    quantity_vec_size = quantity.size()
     n_samples = None
     sums = None
+    sums_power = None
     i_chunk = 0
+
     while True:
```
jbrezmorf: Try to move the condition to the while head, to improve readability.

jbrezmorf: Seems that it is not possible without extracting all chunks at once. Put there a comment, that loop over level is intentional to minimize the number of chunks processed in parallel in order to have a smaller memory footprint.
I.e. truly parallel processing may be beneficial.

src/mlmc/quantity_concept.py
```diff
@@ -1,109 +1,313 @@
...
+            break
+        i_chunk += 1
+
     mean = np.zeros_like(sums[0])
-    for s, n in zip(sums, n_samples):
+    mean_power = np.zeros_like(sums[0])
+    for s, sp, n in zip(sums, sums_power, n_samples):
         mean += s / n
-    return quantity.make_value(mean)
+        mean_power += sp / n
```
jbrezmorf: rename xyz_power to xyz_square

src/mlmc/quantity_concept.py
```diff
@@ -1,109 +1,313 @@
...
+def moment(quantity, moments_fn, i=0):
+    """
+    Create quantity with operation that evaluates particular moment
+    :param quantity: Quantity instance
+    :param moments_fn: mlmc.moments.Moments child
+    :param i: index of moment
+    :return: Quantity
+    """
+    def eval_moment(x):
+        return moments_fn.eval_fine_coarse(i, value=x)
```
jbrezmorf: I see. So rename the method to something like `eval_single_moment`. Put here a note that current implementation is inefficient for large `i`.

src/mlmc/quantity_concept.py
```diff
@@ -1,109 +1,313 @@
...
+        self._input_quantities = input_quantities
+        # List of quantities on which the 'self' depends, their number have to match number of arguments
+        # to the operation.
 
-        # TODO: design a separate class and mechanisms to avoid repetitive evaluation of quantities
-        # that are parents to more then one quantity
-        # can possibly omit memoization of quantities that are just subsets of their own parent
-        #
-        # can e.g. be done through a decorator
+    def storage_id(self):
```
jbrezmorf: It is used only in a single internal check, see the note there.

src/mlmc/quantity_concept.py
```diff
@@ -1,109 +1,313 @@
...
+        return self.qtype.size()
 
+    def get_cache_key(self, level_id, i_chunk):
+        """
+        Create cache key
+        :param level_id: int
+        :param i_chunk: int
+        :return: tuple
+        """
+        return (level_id, i_chunk, id(self), *[id(q) for q in self._input_quantities], id(self._operation)) # parentheses needed due to py36, py37
```
jbrezmorf: Overcomplicated. `(level_id, i_chunk, id(self))` should be enough, as other values are fixed for a single Quantity object. As an optimization we may consider value based caching:

`(level_id, i_chunk, *[q.get_cache_key(...) for q in self._input_quantities], self._operation)`

src/mlmc/quantity_concept.py
```diff
@@ -1,109 +1,313 @@
...
 
-    def _samples(self, i_chunk):
+    @cached(custom_key_maker=get_cache_key)
```
jbrezmorf: We should free the cached values in the `estimate_mean` after we obtain the chunk from the quantity. In fact memoize is possibly to generic for our purpose, the last calculated chunk can be stored in every Quantity instance.

src/mlmc/quantity_concept.py
```diff
@@ -1,109 +1,313 @@
...
+    def samples(self, level_id, i_chunk):
         """
         Yields list of sample chunks for individual levels.
```
jbrezmorf: Fix description.

src/mlmc/quantity_concept.py
```diff
@@ -1,109 +1,313 @@
...
-        This prevents some coies in concatenation.
+        Possibly calls underlying quantities.
+        :param level_id: int
+        :param i_chunk: int
+        :return: np.ndarray
         """
-        chunks_quantity_level = [q._samples(i_chunk) for  q in self._input_quantities]
+        chunks_quantity_level = [q.samples(level_id, i_chunk) for q in self._input_quantities]
+
         is_valid = (ch is not None for ch in chunks_quantity_level)
```
jbrezmorf: rename to is_invalid

src/mlmc/quantity_concept.py
```diff
@@ -1,109 +1,313 @@
...
             assert (all(is_valid))
-            chunks_level_quantity = zip(*chunks_quantity_level)
-            return [self.operation(*one_level_chunks) for one_level_chunks in chunks_level_quantity]
+            # Operation not set return first quantity samples - used in make_root_quantity
+            if self._operation is None:
+                return chunks_quantity_level[0]
+
+            try:  # numpy function does not have __code__
+                # We need to pass level id and chunk id to select operation
+                if all(par in self._operation.__code__.co_varnames for par in ['level_id', 'i_chunk']):
```
jbrezmorf: This hack is not necessary. What about derived QuantitySelection and calling `self._operation` through e.g. `self.call_operation` which can be overloaded there. Python also provides high level introspection tools like `inspect.signature`

src/mlmc/quantity_concept.py
```diff
@@ -1,109 +1,313 @@
...
+            except:
+                pass
+
+            return self._operation(*chunks_quantity_level)
         else:
             return None
 
-
-    def _make_value(self, data_row: np.array):
+    def _make_value(self, mean: np.ndarray, var: np.ndarray):
```
jbrezmorf: Possibly rename to `_create_quantity_mean`.

src/mlmc/quantity_concept.py
```diff
@@ -1,109 +1,313 @@
...
+        assert all(q.size() == quantities[0].size() for q in quantities), "Quantity must have same structure"
+        return Quantity(quantities[0].qtype, operation=operation, input_quantities=quantities)
+
+    def select(self, *args):
+        """
+        Performs sample selection based on conditions
+        :param args: Quantity
+        :return: Quantity
+        """
+        masks = args[0]
```
jbrezmorf: Check len(args).

src/mlmc/quantity_concept.py
```diff
@@ -1,109 +1,313 @@
...
+            try:  # numpy function does not have __code__
+                # We need to pass level id and chunk id to select operation
+                if all(par in self._operation.__code__.co_varnames for par in ['level_id', 'i_chunk']):
+                    return self._operation(*chunks_quantity_level, level_id=level_id, i_chunk=i_chunk)
+            except:
+                pass
+
+            return self._operation(*chunks_quantity_level)
         else:
             return None
```
jbrezmorf: Should we raise an exception instead of returning None? there should be no None among the input chunks.

src/mlmc/quantity_concept.py
```diff
@@ -1,109 +1,313 @@
...
+            "All mask quantities must have base type BoolType"
+
+        # More conditions leads to default AND
+        if len(args) > 1:
+            for m in args[1:]:
+                masks = logical_and(masks, m)  # method from this module
 
+        def op(x, level_id, i_chunk):
+            mask = masks.samples(i_chunk, level_id)
+            if mask is not None:
```
jbrezmorf: None should not happen.

src/mlmc/quantity_concept.py
```diff
@@ -121,120 +325,451 @@ def mult_op(x, y):
     def __rmul__(self, other):
```
jbrezmorf: What about other operators?

src/mlmc/quantity_concept.py
```diff
@@ -121,120 +325,451 @@ def mult_op(x, y):
...
+        :param x: Quantity
+        :param y: Quantity or int, float
+        :param operator: operator module function
+        :return: np.ndarray of bools
+        """
+        # all values for sample must meet given condition,
+        # if any value doesn't meet the condition, whole sample is eliminated
+
+        # It is most likely zero level, so use just fine samples
+        if np.all((x[..., 1] == 0)):
```
jbrezmorf: We should document this behavior somewhere, possibly there should be a comprehensive description of the Quantity class.

jbrezmorf: Detection of the zero level in this way is incorrect. No way we must propagate the level information with the chunk array.

src/mlmc/quantity_concept.py
```diff
@@ -121,120 +325,451 @@ def mult_op(x, y):
...
+        """
+        Get items from Quantity, quantity type must support brackets access
+        :param key: str, int, tuple
+        :return: Quantity
+        """
+        new_qtype = self.qtype[key]  # New quantity type
+
+        reshape_shape = None
+        # ArrayType might be accessed directly regardless of qtype start and size
+        if isinstance(self.qtype, ArrayType):
```
jbrezmorf: Try to avoid usage of isinstance, use overloaded method instead.

Here we can have key acces implemented in the QType, e.g.:

    return Quantity(quantity_type=new_qtype, input_quantities=[self],
         operation=self.qtype._make_getitem_op(...))

src/mlmc/quantity_concept.py
```diff
@@ -121,120 +325,451 @@ def mult_op(x, y):
...
+    def __copy__(self):
+        return Quantity(quantity_type=self.qtype, input_quantities=custom_copy(self._input_quantities),
+                        operation=self._operation)
 
-        self.time_frames = time_frames
-        # list of pairs (time, value)
-        self.block_starts = []
-        # starts of blocks of the time values on the single sample row
-        self.operation = lambda lch : np.concatenate(lch, axis=0)
+    def time_interpolation(self, value):
```
jbrezmorf: I'm not sure this would work for any Quantity type, `self.qtype_times` may be invalid. That should probably be specific to the TimeSeries type.

src/mlmc/quantity_concept.py
```diff
@@ -121,120 +325,451 @@ def mult_op(x, y):
...
+        return self._var
 
-        def max_op(*input_chunks):
-            table = np.stack(input_chunks)
-            return np.max(table, axis=0)
+    def mean(self):
+        return self._mean
 
-        return self.frame0._copy(*[q for t, q in self.time_frames], operation=max_op)
+    def __getitem__(self, key):
```
jbrezmorf: Should be reused from quantity or rather from QType descendants. See the note above.

src/mlmc/quantity_concept.py
```diff
@@ -121,120 +325,451 @@ def mult_op(x, y):
...
+        """
+        ArrayType indexing
+        :param key: int, tuple of ints or slice objects
+        :return: QuantityType - ArrayType or self._qtype
+        """
+        # Get new shape
+        new_shape = np.empty(self._shape)[key].shape
+
+        # One selected item is considered to be a scalar QType
+        if len(new_shape) == 1 and new_shape[0] == 1:
```
jbrezmorf: We should keep the produced new_shape. Be as close to the numpy behavior as possible.

src/mlmc/quantity_concept.py
```diff
@@ -121,120 +325,451 @@ def mult_op(x, y):
...
+            new_shape = ()
+
+        # Result is also array
+        if len(new_shape) > 0:
+            q_type = ArrayType(new_shape, qtype=copy.deepcopy(self._qtype))
```
jbrezmorf: Why we need deep copy of the type?

src/mlmc/quantity_concept.py
```diff
@@ -121,120 +325,451 @@ def mult_op(x, y):
...
 
+        q_type.start = size
 
+        return q_type
 
-class Field(QBase):
-    def __init__(self, *args: Tuple[str, QBase]):
-        self.quantity_dict = OrderedDict(*args)
 
+def custom_copy(quantities):
```
jbrezmorf: Possibly make it a private static method of Quantity and use appropriate name. 'custom' says nothing.

test/test_quantity_concept.py
```diff
@@ -0,0 +1,468 @@
+import os
```
jbrezmorf: I  have to go through the test yet...

jbrezmorf (commented):
Well done. Just some minor points for improvement.

PR #105, opened 2020-11-03
Jb review quantity

jbrezmorf:


src/mlmc/moments.py
```diff
@@ -6,12 +6,13 @@ class Moments:
     """
     Class for moments of random distribution
     """
-    def __init__(self, size, domain, log=False, safe_eval=True):
+    def __init__(self, size, domain, log=False, safe_eval=True, mean=0):
```
jbrezmorf, line 9: In order to keep basic moment classes simple we can make moment centering and possibly rescaling through quantities. Moreover the centered moments can be (with some numerical error) computed from noncentered. We should test the numerical error for the real data.

src/mlmc/quantity_concept.py
```diff
@@ -1,115 +1,430 @@
+import abc
 import numpy as np
+import copy
```
jbrezmorf, line 3: The file should be splitted into more smaller files. Partially done in subsequent commits.
Proposed files:
- quantity_type
- quantity_estimate
- quantity

src/mlmc/quantity_concept.py
```diff
@@ -121,120 +436,583 @@ def mult_op(x, y):
...
+        def op_concatenate(*chunks):
+            y = np.concatenate(tuple(chunks), axis=axis)
+            return y
+        return Quantity(qtype, input_quantities=[*quantities], operation=op_concatenate)
+
+
+class QuantityConst(Quantity):
+    def __init__(self, quantity_type, values, input_quantities=None, operation=None):
+        """
+        Quantity class represents real quantity and also provides operation that can be performed with stored values.
```
jbrezmorf: Fix documentation particular for the QuantityConst.

src/mlmc/quantity_concept.py
```diff
@@ -1,115 +1,430 @@
...
+        :param i_chunk: int
+        :return: np.ndarray
         """
-        chunks_quantity_level = [q._samples(i_chunk) for  q in self._input_quantities]
+        if not all(np.allclose(q.storage_id(), self._input_quantities[0].storage_id()) for q in self._input_quantities):
+            raise Exception("Not all input quantities come from the same quantity storage")
+
+        chunks_quantity_level = [q.samples(level_id, i_chunk) for q in self._input_quantities]
+
         is_valid = (ch is not None for ch in chunks_quantity_level)
```
jbrezmorf: Can the samples function return None ever? We should possibly convert this into an assert.

jbrezmorf: But on single lin, since asserts can be turned off in a production run.

src/mlmc/quantity_concept.py
```diff
@@ -1,115 +1,430 @@
...
         if any(is_valid):
             assert (all(is_valid))
-            chunks_level_quantity = zip(*chunks_quantity_level)
-            return [self.operation(*one_level_chunks) for one_level_chunks in chunks_level_quantity]
+            # Operation not set return first quantity samples - used in make_root_quantity
+            if self._operation is None:
```
jbrezmorf: Can this happen?

src/mlmc/quantity_concept.py
```diff
@@ -1,115 +1,430 @@
...
+                return chunks_quantity_level[0]
+
+            additional_params = {}
+            try:
```
jbrezmorf: This can be done once out of the computational intensive operation.
Anyway, which operations need level_id and i_chunk? This should not happen.

src/mlmc/quantity_concept.py
```diff
@@ -121,120 +436,583 @@ def mult_op(x, y):
...
         """
-        def max_op(*input_chunks):
-            table = np.stack(input_chunks)
-            return np.max(table, axis=0)
+        #@TODO: try to use method from Quantity class
+        return (level_id, i_chunk, id(self), n_samples)  # redundant parentheses needed due to py36, py37
 
-        return self.frame0._copy(*[q for t, q in self.time_frames], operation = max_op)
+    @cached(custom_key_maker=get_cache_key)
+    def samples(self, level_id, i_chunk, n_samples=np.inf):
```
jbrezmorf, line 660: Rename to _samples as it is an internal method. 
Avoid special arguments (np.inf, None) in order to simplify the code, we have control over calling the interal methods.

src/mlmc/quantity_concept.py
```diff
@@ -121,120 +436,583 @@ def mult_op(x, y):
...
+
+        # Get first n samples
+        if n_samples is not None and n_samples < np.inf:
+            return chunks_quantity_level[0][:, :n_samples, :]
+
+        items_in_chunk = quantity_storage._storage.get_items_in_chunk(level_id)
+        if items_in_chunk is None:
+            quantity_storage.samples(level_id, i_chunk, n_samples)
+
+        items_in_chunk = quantity_storage._storage.get_items_in_chunk(level_id)
```
jbrezmorf: rename `get_items_in_chunk` e.g. to `level_n_samples`.

src/mlmc/quantity_concept.py
```diff
@@ -121,120 +436,583 @@ def mult_op(x, y):
...
+            arr[:] = self._values
+            self._values = arr
+        chunks_quantity_level = [self._values]
+
+        # Get first n samples
+        if n_samples is not None and n_samples < np.inf:
+            return chunks_quantity_level[0][:, :n_samples, :]
+
+        items_in_chunk = quantity_storage._storage.get_items_in_chunk(level_id)
+        if items_in_chunk is None:
```
jbrezmorf: What is the point of this? Which case this should cover?

src/mlmc/quantity_concept.py
```diff
@@ -121,120 +436,583 @@ def mult_op(x, y):
...
+            quantity_storage.samples(level_id, i_chunk, n_samples)
+
+        items_in_chunk = quantity_storage._storage.get_items_in_chunk(level_id)
+
+        if items_in_chunk is None:
+            return None
```
jbrezmorf: Avoid retuning None.

src/mlmc/quantity_concept.py
```diff
@@ -121,120 +436,583 @@ def mult_op(x, y):
...
+
+        level_chunk = self._values[:, i_chunk * items_in_chunk: (i_chunk + 1) * items_in_chunk, :]
```
jbrezmorf: This seems like _values containing the samples, that should not be allowed. Samples must com only from the SampleStorage.

src/mlmc/quantity_concept.py
```diff
@@ -121,120 +436,583 @@ def mult_op(x, y):
...
+        assert self.qtype.size() == level_chunk.shape[0]
+
+        if level_chunk.shape[1] == 0:
+            return None
 
-    def interpolate(self):
+        # Select values from given interval self.start:self.end
+        if quantity_storage.start is not None and quantity_storage.end is not None:
```
jbrezmorf: Rest is meaningless, QuantityConstant should never perform an operation.

src/mlmc/quantity_concept.py
```diff
@@ -1,115 +1,430 @@
...
+    if isinstance(result, (int, float, bool)):
+        qtype = base_qtype_val
+    elif isinstance(result, (list, np.ndarray)):
+        result = np.array(result)
+        qtype = ArrayType(shape=result.shape[0], qtype=base_qtype_val)
+
+    return qtype
+
+
+def _create_quantity(values, base_qtype, quantity_storage):
```
jbrezmorf: Convert to the class method `Quantity.wrap(x)`.
- Do we need to pass base_qtype? The Bool base type should be only result of quantity operations.
- We needn't pass storage either as the samples method have n_samples argument. The point of introducing something like that is to keep QuantityConstant independent of the storage.
- But it turns out that we need the storage, we can pass it to the wrap method.

src/mlmc/quantity_concept.py
```diff
@@ -1,115 +1,430 @@
...
+import operator
+from inspect import signature
+from memoization import cached
+from scipy import interpolate
+from typing import List, Tuple
+from mlmc.sample_storage import SampleStorage
+from mlmc.sim.simulation import QuantitySpec
+
+
+def _get_quantity_info(args_quantities, get_quantity_storage=False):
```
jbrezmorf: - Separate into two functions.
- make these the class methods of the quantity together with the _method function

src/mlmc/quantity_concept.py
```diff
@@ -1,115 +1,430 @@
...
+    Process input parameters to perform numpy ufunc.
+    Get base QType of passed quantities, QuantitiStorage instance, ...
+    Determine the resulting QType from the first few samples
+    :param ufunc: ufunc object that was called
+    :param method: string, indicating which Ufunc method was called
+    :param args: tuple of the input arguments to the ufunc
+    :param kwargs: dictionary containing the optional input arguments of the ufunc
+    :return: Quantity
+    """
+    def _demo_method(*input_quantities_chunks):
```
jbrezmorf: - rename, e.g. ufunc_call
- no need to treat single arg case diferently

src/mlmc/quantity_concept.py
```diff
@@ -1,115 +1,430 @@
...
+        if len(input_quantities_chunks) == 1:
+            return getattr(ufunc, method)(input_quantities_chunks[0], **kwargs)
+        else:
+            return getattr(ufunc, method)(*input_quantities_chunks, **kwargs)
+
+    base_qtype, quantity_storage = _get_quantity_info(args, get_quantity_storage=True)
+
+    quantities = []
+    for arg in args:
+        if not isinstance(arg, Quantity):
```
jbrezmorf: This `if` should be part of the wrap function.

src/mlmc/quantity_concept.py
```diff
@@ -1,115 +1,430 @@
...
+            quantities.append(_create_quantity(arg, base_qtype=base_qtype, quantity_storage=quantity_storage))
+        else:
+            quantities.append(arg)
+
+    # Check if all quantities come from same storage
+    if not all(np.allclose(q.storage_id(), quantities[0].storage_id()) for q in quantities[1:]):
```
jbrezmorf: This check should probably be part of the Quantity constructor.

src/mlmc/quantity_concept.py
```diff
@@ -1,115 +1,430 @@
...
+        if isinstance(quantity, Quantity):
+            if get_quantity_storage:
+                q_storage = quantity.get_quantity_storage()
+            if type(quantity.qtype.base_qtype()) == ScalarType:
+                return ScalarType(), q_storage
+
+    return BoolType(), q_storage
+
+
+def _determine_qtype(quantities, method):
```
jbrezmorf: Rename `_result_qtype`, method should be rather the first parameter.

src/mlmc/quantity_concept.py
```diff
@@ -1,115 +1,430 @@
...
+    """
+    Determine QType from evaluation with given method and first few samples from storage
+    :param quantities: list of Quantities
+    :param method: ufunc function
+    :return: QType
+    """
+    chunks_quantity_level = [q.samples(level_id=0, i_chunk=0, n_samples=10) for q in quantities]
```
jbrezmorf, line 22: This complicates the samples method, what about to introduce a separate method that gets just few samples for level 0. Would help if we can separate the recursion mechanism from the actual recursive call. We have to start with the simplification of the samples method.

src/mlmc/quantity_concept.py
```diff
@@ -1,115 +1,430 @@
...
+    result = method(*chunks_quantity_level)
+
+    base_qtype_val, _ = _get_quantity_info(quantities)
+
+    if isinstance(result, (int, float, bool)):
```
jbrezmorf: We ask for 10 samples so the result should always be an array so this is possibly wrong.

src/mlmc/quantity_concept.py
```diff
@@ -1,115 +1,430 @@
...
+                if level_id == 0:
+                    # Set variables for level sums and sums of powers
+                    if i_chunk == 0:
+                        sums = [np.zeros(chunk.shape[0]) for _ in range(n_levels)]
+                        sums_of_squares = [np.zeros(chunk.shape[0]) for _ in range(n_levels)]
+
+                    # Coarse result for level 0, there is issue for moments processing (not know about level)
+                    chunk[..., 1] = 0
+
+                chunk = remove_nan_samples(chunk)
```
jbrezmorf: We should be able to report number of removed samples.

src/mlmc/quantity_concept.py
```diff
@@ -1,115 +1,430 @@
...
+        array_type = ArrayType(q_spec.shape, scalar_type)
+        field_type = FieldType([(loc, array_type) for loc in q_spec.locations])
+        ts_type = TimeSeriesType(q_spec.times, field_type)
+        dict_types.append((q_spec.name, ts_type))
+    dict_type = DictType(dict_types)
+
+    return QuantityStorage(storage, dict_type)
+
+
+def remove_nan_samples(chunk):
```
jbrezmorf: Rename to `mask_nan_samples`.

src/mlmc/quantity_concept.py
```diff
@@ -1,115 +1,430 @@
...
+    Create a root quantity that has QuantityStorage as the input quantity,
+    QuantityStorage is the only class that directly accesses the stored data.
+    Quantity type is created based on the q_spec parameter
+    :param storage: SampleStorage
+    :param q_specs: same as result format in simulation class
+    :return: QuantityStorage
+    """
+    # Set chunk size as the case may be
+    if storage.chunk_size is None:
+        storage.chunk_size = 512000  # bytes in decimal
```
jbrezmorf: - Make this a named constant. 
- Setting the chunk size to the storage seems inapropriate, possibly we can introduce a small object
ChunkSpec, that could be passed through the samples methods down to the storage and then the storage returns the appropriate chunk. Then the chunk size can be (optional) parameter of the `estimate_mean` which is better since then we can possibly set the chunk_size dinamically according to the free memory or cache size ...
Moreover this can be used to get try chunk for determining the result type of the ufunc.

jbrezmorf (commented):
Just trying to start somewhere with this enourmous review.
- can not review this at once because of the big file and lot of interconnected problems.
- this review focuss mainly to the treating of the ufuncs and samples methods

src/mlmc/moments.py
jbrezmorf: separate issue

src/mlmc/quantity_concept.py
```diff
@@ -1,115 +1,430 @@
...
+    Convert flat, bool or array (list) to Quantity
+    :param values: flat, bool, array (list)
+    :param base_qtype: QType
+    :return: Quantity
+    """
+    if isinstance(values, (int, float, bool)):
+        quantity = QuantityConst(quantity_type=base_qtype, values=values, input_quantities=[quantity_storage])
+    elif isinstance(values, (list, np.ndarray)):
+        values = np.array(values)
+        qtype = ArrayType(shape=values.shape[0], qtype=base_qtype)
```
jbrezmorf: `shape=values.shape`

src/mlmc/quantity_concept.py
jbrezmorf: In order to relax passing storage to QuantityConstant we must modify the test for the same storage of all quantities in single calculation.

src/mlmc/quantity_concept.py
jbrezmorf: None currently indicates end of iteration over the storage. We can replace this by raising the StopIteration exception.

src/mlmc/quantity_concept.py
jbrezmorf: level_id and i_chunk is used by mask operation and bootstrapping, need a review why it is necessary.

src/mlmc/quantity_concept.py
martinspetlik: separate issue created

src/mlmc/quantity_concept.py
martinspetlik: separate issue created

src/mlmc/quantity_concept.py
martinspetlik: After a few changes, it is not necessary to get quantity storage.

src/mlmc/quantity_concept.py
martinspetlik: separate issue created

src/mlmc/quantity_concept.py
```diff
@@ -1,115 +1,367 @@
...
-        for i_level, level_chunk in enumerate(chunk):
-            # level_chunk is Numpy Array with shape [M, chunk_size, 2]
-            n_samples[i_level] += len(level_chunk)
-            assert(level_chunk.shape[1] == quantity_vec_size)
-            sums[i_level] += np.sum(level_chunk[:, :, 0] - level_chunk[:, :, 1], axis=1)
+
+        level_chunks_none = np.zeros(n_levels)
+        for level_id in level_ids:
+            # Chunk of samples for given level id
+            try:
```
jbrezmorf, line 92: Missing loop over chunks. With ChunkSpec we can change the `samples` method to a generator.

src/mlmc/quantity_concept.py
```diff
@@ -1,115 +1,367 @@
...
+
+                    # Coarse result for level 0, there is issue for moments processing (not know about level)
+                    chunk[..., 1] = 0
+
+                chunk = mask_nan_samples(chunk)
+                # level_chunk is Numpy Array with shape [M, chunk_size, 2]
+                n_samples[level_id] += chunk.shape[1]
+
+                assert(chunk.shape[0] == quantity_vec_size)
+                sums[level_id] += np.sum(chunk[:, :, 0] - chunk[:, :, 1], axis=1)
```
jbrezmorf: Compute difference `chunk[:, :, 0] - chunk[:, :, 1]` to a variable and use it twice.

src/mlmc/quantity_concept.py
```diff
@@ -1,115 +1,367 @@
...
+    def __init__(self, quantity_type, input_quantities=[], operation=None):
+        """
+        Quantity class represents real quantity and also provides operation that can be performed with stored values.
+        Each Quantity has Qtype which describes its structure.
+        :param quantity_type: QType instance
+        :param input_quantities: List[Quantity]
+        :param operation: function
+        """
+        self.qtype = quantity_type
         self._operation = operation
```
jbrezmorf, line 207: move `operation` before `input_quantities` and make it obligatory, None value should be invalid in Quantity.

src/mlmc/quantity_concept.py
```diff
@@ -1,115 +1,367 @@
...
-        else:
-            return None
+        Possibly calls underlying quantities.
+        :param level_id: int
+        :param i_chunk: int
+        :return: np.ndarray
+        """
+        chunks_quantity_level = [q.samples(level_id, i_chunk) for q in self._input_quantities]
+
+        if self._operation is None:
```
jbrezmorf: None must be invalid and the returned value would not be valid anyway. So remove this test.

src/mlmc/quantity_concept.py
```diff
@@ -1,115 +1,367 @@
...
 
-        # TODO: design a separate class and mechanisms to avoid repetitive evaluation of quantities
-        # that are parents to more then one quantity
-        # can possibly omit memoization of quantities that are just subsets of their own parent
-        #
-        # can e.g. be done through a decorator
+        self._check_storage_ids()
+        self._op_additional_params()
+
+    def _check_storage_ids(self):
```
jbrezmorf: Since `_check_storage_ids` is performed in the constructor we are sure there is one single storege_id and that should be remembered and returned by `storage_id` methods. that can further simplify the implementation of this method.

src/mlmc/quantity_concept.py
```diff
@@ -1,115 +1,367 @@
...
+        """
+        st_ids = []
+        if len(self._input_quantities) == 0:
+            return None
+        for input_quantity in self._input_quantities:
+            st_id = input_quantity.storage_id()
+            st_ids.extend(st_id if type(st_id) == list else [st_id])
+        return st_ids
 
+    def get_quantity_storage(self):
```
jbrezmorf: If we need this function then we can check and save (reference to) the common storage in `_check_storage_id` then storage_id just returns `id(self._storage)`

src/mlmc/quantity_concept.py
```diff
@@ -1,115 +1,367 @@
...
+        return self._input_quantities[0].get_quantity_storage()
 
-    def _samples(self, i_chunk):
+    def size(self) -> int:
```
jbrezmorf, line 271: Should we make this a private method?

src/mlmc/quantity_concept.py
```diff
@@ -1,115 +1,367 @@
...
         Crate a new quantity with the same structure but containing fixed data vector.
         Primary usage is to organise computed means and variances.
         Can possibly be used also to organise single sample row.
-        :param data_row:
+        :param mean: np.ndarray
+        :param var: np.ndarray
         :return:
         """
-        pass
+        if np.isnan(mean).all():
```
jbrezmorf, line 313: Is this test meaningfull, is that case somehow special? If not we can just call the `QuantityMean` constructor and remove this method.

src/mlmc/quantity_concept.py
```diff
@@ -1,115 +1,367 @@
...
+        Performs sample selection based on conditions
+        :param args: Quantity
+        :return: Quantity
+        """
+        # args always has len() at least 1
+        masks = args[0]
+
+        for quantity in args:
+            if not isinstance(quantity.qtype.base_qtype(), BoolType):
+                raise Exception("Quantity: {} doesn't have BoolType, instead it has QType: {}"
```
jbrezmorf, line 341: What happens if I do the same with the `logical_and` directly? I.e.

`q_and = np.logical_and(True, root_quantity)`

that should report some meaningful error similar to this exception.

src/mlmc/quantity_concept.py
```diff
@@ -1,115 +1,367 @@
...
+                                .format(quantity, quantity.qtype.base_qtype()))
+
+        # More conditions leads to default AND
+        if len(args) > 1:
+            for m in args[1:]:
+                masks = np.logical_and(masks, m)  # method from this module
 
+        def op(x, level_id, i_chunk):
```
jbrezmorf: We can possibly get rid of `level_id` and `chunk_id` optional parameters as follows:
```
def mask_op(x, mask):
    return x[..., mask, :]

return Quantity(quantity_type=self.qtype, input_quantities=[self, masks], operation=mask_op)
```
Nevertheless, we must force `storage_id` of the created quantity to be defferent inorder to not mix quantities with different number of samples (different mask). Therefore we should rename `storeage_id` to e.g. `selection_id`, but still take it as the `id(storage)` for the root quantity.

src/mlmc/quantity_concept.py
```diff
@@ -1,115 +1,367 @@
...
+        return Quantity._method(ufunc, method, *args, **kwargs)
 
     def __add__(self, other):
         def add_op(x, y):
             return x + y
         return self._reduction_op([self, other], add_op)
 
+    def __sub__(self, other):
+        def sub_op(x, y):
+            return x + y
```
jbrezmorf: error

src/mlmc/quantity_concept.py
```diff
@@ -1,115 +1,367 @@
...
+
+    def __mul__(self, other):
+        if isinstance(other, (float, int)):
+            return self.__const_mult(other)
+
+        def mult_op(x, y):
+            return x * y
+        return self._reduction_op([self, other], mult_op)
+
     def __mul__(self, other):
```
jbrezmorf: So this should be removed?

src/mlmc/quantity_concept.py
```diff
@@ -121,120 +373,594 @@ def mult_op(x, y):
     def __rmul__(self, other):
         if isinstance(other, (float, int)):
             return self.__const_mult(other)
-        assert(False)
+        assert False
 
     def __const_mult(self, other):
```
jbrezmorf: Add at least: `__truediv__`, `__mod__`
possibly [others](https://diveintopython3.net/special-method-names.html).

Add all complementary __r variants.

Comparisons are [OK](https://docs.python.org/3/reference/datamodel.html#object.__eq__).

src/mlmc/quantity_concept.py
```diff
@@ -121,120 +373,594 @@ def mult_op(x, y):
...
+        """
+        bool_type = BoolType()
+        new_qtype = copy.deepcopy(self.qtype)
+        new_qtype.replace_scalar(bool_type)
 
-    def set(self, range_begin, size):
-        self.begin = range_begin
-        self._size = size
-        self.end = range_begin + size
+        if isinstance(other, (float, int)):
```
jbrezmorf: Just use `Quantity.wrap` here and in other operators.

src/mlmc/quantity_concept.py
```diff
@@ -121,120 +373,594 @@ def mult_op(x, y):
...
+            if not isinstance(self.qtype.base_qtype(), ScalarType):
+                raise TypeError("Quantity has base qtype {}. "
+                                "Quantities with base qtype ScalarType are the only ones that support comparison".
+                                format(self.qtype.base_qtype()))
+
+            return Quantity(quantity_type=new_qtype, input_quantities=[self], operation=op)
```
jbrezmorf: This would not work since `op` expects two arguments.

src/mlmc/quantity_concept.py
```diff
@@ -121,120 +373,594 @@ def mult_op(x, y):
...
+        elif isinstance(other, Quantity):
+            if self.qtype.size() != other.qtype.size():
+                raise Exception("Quantities don't have have the same size")
+            if not isinstance(self.qtype.base_qtype(), ScalarType) or not isinstance(other.qtype.base_qtype(), ScalarType):
```
jbrezmorf: If we do not check this in other operations we should test it here neither. Or we should test it in the Quantity constructor and only suppress it when applying the mask.

jbrezmorf: The previous check for size already implemented in `_reduce_op` so we can call it.

src/mlmc/quantity_concept.py
```diff
@@ -121,120 +373,594 @@ def mult_op(x, y):
...
+        whole sample is eliminated
+        :param x: Quantity
+        :param y: Quantity or int, float
+        :param operator: operator module function
+        :param level_id: int, level identifier
+        :return: np.ndarray of bools
+        """
+        # Zero level - use just fine samples
+        if level_id == 0:
+            if isinstance(y, int) or isinstance(y, float):
```
jbrezmorf: Handle this by `Quantity.wrap`

src/mlmc/quantity_concept.py
```diff
@@ -121,120 +373,594 @@ def mult_op(x, y):
...
+        All values for sample must meet given condition, if any value doesn't meet the condition,
+        whole sample is eliminated
+        :param x: Quantity
+        :param y: Quantity or int, float
+        :param operator: operator module function
+        :param level_id: int, level identifier
+        :return: np.ndarray of bools
+        """
+        # Zero level - use just fine samples
+        if level_id == 0:
```
jbrezmorf, line 421: Can we just reduce shape to (M,1,N) for level 0 starting from the storage and get rid of this special case?

src/mlmc/quantity_concept.py
```diff
@@ -121,120 +373,594 @@ def mult_op(x, y):
...
+    def __eq__(self, other):
+        def eq_op(x, y=other, level_id=0):
+            return self._process_mask(x, y, operator.eq, level_id)
+        return self._mask_quantity(other, eq_op)
+
+    def __ne__(self, other):
+        def ne_op(x, y=other, level_id=0):
+            return self._process_mask(x, y, operator.ne, level_id)
+        return self._mask_quantity(other, ne_op)
+
```
jbrezmorf, line 475: CONTINUE HERE.

jbrezmorf (commented):
Partial progress in the review.
I will continue.

src/mlmc/quantity_concept.py
```diff
@@ -121,120 +373,594 @@ def mult_op(x, y):
...
+    def sampling(self, size):
+        """
+        Random sampling
+        :param size: number of samples
+        :return: np.ndarray
+        """
+        def mask_gen(x, *args):
```
jbrezmorf, line 482: This is not so simple. In order to allow bootstrapping we need:
1. set the subset size for every level individually
2. level sample size must be the size accross all chunks not only from the one.
The simplest way to implement this is possibly to make a descendant of Quantity and overwrite the sampling method. We can use `self._storage` and/or ChunkSpec.
In particular we have to know total numer of samples on the actual level and return a subset of the size `target_size*chunk_size/level_size`. Some stipulations must be done to have exactly `target_size` samples in total.

src/mlmc/quantity_concept.py
```diff
@@ -121,120 +373,594 @@ def mult_op(x, y):
...
+            indices = np.random.choice(x.shape[1], size=size)
+            mask = np.zeros(x.shape[1], bool)
+            mask[indices] = True
+            return mask
+        return self._mask_quantity(size, mask_gen)
+
+    def __getitem__(self, key):
```
jbrezmorf, line 489: error reporting issue #117

src/mlmc/quantity_concept.py
```diff
@@ -121,120 +373,594 @@ def mult_op(x, y):
...
+        Operation
+        :param chunk: level chunk, list with shape [M, chunk size, 2]
+        :param new_qtype: QType
+        :param key: parent QType's key, needed for ArrayType
+        :return: list
+        """
+        start = new_qtype.start
+        end = new_qtype.start + new_qtype.size()
+        slice_key = slice(start, end)
+        return self._keep_dims(chunk[slice_key])
```
jbrezmorf: The shape should be preserved, so no need to call `_keep_dims`.

src/mlmc/quantity_concept.py
```diff
@@ -121,120 +373,594 @@ def mult_op(x, y):
...
 
-    def _samples(self, i_chunk):
-        level_chunks = self.storage.sample_pairs(i_chunk)
-        return [chunk[self.begin:self.end, :, :]  for chunk in level_chunks]
+        def _make_getitem_op(y):
+            return self.qtype._make_getitem_op(y, new_qtype=new_qtype, key=key)
 
+        return Quantity(quantity_type=new_qtype, input_quantities=[self], operation=_make_getitem_op)
 
+    def __iter__(self):
```
jbrezmorf: Just drop this method and let Python do the job.

src/mlmc/quantity_concept.py
```diff
@@ -121,120 +373,594 @@ def mult_op(x, y):
...
+        raise Exception("This class is not iterable")
 
+    def __copy__(self):
```
jbrezmorf: Why we need the shallow copy? I have found only copies of qtype.

src/mlmc/quantity_concept.py
```diff
@@ -121,120 +373,594 @@ def mult_op(x, y):
...
-    """
-    We should assume that all quantities have same structure (check only same size)
-    but can be of any type. Meanigful are only onther types then TimeSeries.
-    """
-    def __init__(self, time_frames : List[Tuple[float, QBase]]):
-        super().__init__(time_frames)
-        self.frame0 = time_frames[0][1]
-        assert( all( (q._size == self.frame0._size for t,q in time_frames) ) )
-        assert( all( (q.__class__ == self.frame0.__class__ for t,q in time_frames) ))
+    def time_interpolation(self, value):
```
jbrezmorf: `time_interpolation` is only supported by the TimeSeries qtype. This is currently the only qtype specific method, but there will be more of these later on. To this and we should have general __getattr__ with qtype resolution like:
```
def __getattr__(self, name):
   static_fun = getattr(self._qtype, name) #we support only static function call forwarding
   def apply_on_quantity(*attr, **d_attr):
      return fun(self, *attr, **d_attr)
   return apply_on_qunatity
```

jbrezmorf: Then we can move the implementation into the TimeSeries qtype.

src/mlmc/quantity_concept.py
```diff
@@ -121,120 +373,594 @@ def mult_op(x, y):
...
+            return f(value)
 
+        return Quantity(quantity_type=self.qtype._qtype, input_quantities=[self], operation=interp)
 
-        self.time_frames = time_frames
-        # list of pairs (time, value)
-        self.block_starts = []
-        # starts of blocks of the time values on the single sample row
-        self.operation = lambda lch : np.concatenate(lch, axis=0)
+    def level_ids(self):
```
jbrezmorf: If we have `self._storage` we can simplify this recursion.

jbrezmorf: But even better is to drop this method and use:
`top_quantity.get_storage().level_ids()` in the `estimate_mean`.

src/mlmc/quantity_concept.py
```diff
@@ -121,120 +373,594 @@ def mult_op(x, y):
...
+        :return: List[int]
+        """
+        if len(self._input_quantities) == 0:
+            return []
+        for q in self._input_quantities:
+            if len(q.level_ids()) > 0:
+                return q.level_ids()
+
+    @staticmethod
+    def concatenate(quantities, qtype, axis=0):
```
jbrezmorf: Yep. Another qtype specific method, supported only by Array qtype. Then result qtype is again Array

src/mlmc/quantity_concept.py
```diff
@@ -121,120 +373,594 @@ def mult_op(x, y):
...
+        :return: Quantity
+        """
+        def op_concatenate(*chunks):
+            y = np.concatenate(tuple(chunks), axis=axis)
+            return y
+        return Quantity(qtype, input_quantities=[*quantities], operation=op_concatenate)
 
-    def max(self):
+    @staticmethod
+    def _get_base_qtype(args_quantities):
```
jbrezmorf, line 527: We should check that the base type is common, shouldn't we?

src/mlmc/quantity_concept.py
```diff
@@ -121,120 +373,594 @@ def mult_op(x, y):
...
+        for arg in args:
+            quantities.append(Quantity.wrap(arg, base_qtype=Quantity._get_base_qtype(args)))
 
-        return self.frame0._copy(*[q for t, q in self.time_frames], operation = max_op)
+        result_qtype = _result_qtype(_ufunc_call, quantities)
+        return Quantity(quantity_type=result_qtype, input_quantities=list(quantities), operation=_ufunc_call)
 
-    def interpolate(self):
+    @staticmethod
+    def wrap(values, base_qtype):
```
jbrezmorf: We can detect base_qtype.

`isinstance(value, bool)`

`value.dtype == bool`

src/mlmc/quantity_concept.py
```diff
@@ -121,120 +373,594 @@ def mult_op(x, y):
...
+
+    @cached(custom_key_maker=get_cache_key)
+    def samples(self, level_id, i_chunk, n_samples=np.inf):
+        """
+        Get constant values with an enlarged number of axes
+        :param level_id: int
+        :param i_chunk: int
+        :return: np.ndarray
+        """
+        if isinstance(self._values, (int, float, bool)):
```
jbrezmorf: We can do this in the constructor as well.

src/mlmc/quantity_concept.py
```diff
@@ -121,120 +373,594 @@ def mult_op(x, y):
...
+        # Other accessible quantity types uses start and size
+        else:
+            start = new_qtype.start
+            end = new_qtype.start + new_qtype.size()
+            slice_key = slice(start, end)
+
+        mean = self._mean
+        var = self._var
+
+        if reshape_shape is not None:
```
jbrezmorf, line 722: We should follow Quantity approach and keep `self._mean` and `self._var` in the shape `[M]`. And do the reshape in the call operator using a qtype dispatch.

jbrezmorf: The point is that there may be more operations on QuantityMean and we have to preserve the data in a single consistent format to operate on them.

src/mlmc/quantity_concept.py
```diff
@@ -121,120 +373,594 @@ def mult_op(x, y):
...
+        """
+        level_chunk = self._storage.sample_pairs_level(level_id, i_chunk, n_samples=n_samples)  # Array[M, chunk size, 2]
+        if level_chunk is not None:
+            assert self.qtype.size() == level_chunk.shape[0]
+            # Select values from given interval self.start:self.end
+            if self.start is not None and self.end is not None:
+                return level_chunk[self.start:self.end, :, :]
+        return level_chunk
+
+    def __copy__(self):
```
jbrezmorf: Do we need the shallow copy?

src/mlmc/quantity_concept.py
```diff
@@ -121,120 +373,594 @@ def mult_op(x, y):
...
+    def samples(self, level_id, i_chunk, n_samples=np.inf):
+        """
+        Get results for given level id and chunk id
+        :param level_id: int
+        :param i_chunk: int
+        :param n_samples: int, number of retrieved samples
+        :return: Array[M, chunk size, 2]
+        """
+        level_chunk = self._storage.sample_pairs_level(level_id, i_chunk, n_samples=n_samples)  # Array[M, chunk size, 2]
+        if level_chunk is not None:
```
jbrezmorf: Rather an assert. We use StopIteration exception.

src/mlmc/quantity_concept.py
```diff
@@ -121,120 +373,594 @@ def mult_op(x, y):
...
+            assert self.qtype.size() == level_chunk.shape[0]
+            # Select values from given interval self.start:self.end
+            if self.start is not None and self.end is not None:
```
jbrezmorf: The root quantity should take whole samples, seems that self.start and self.end are always None.

src/mlmc/quantity_concept.py
```diff
@@ -121,120 +373,594 @@ def mult_op(x, y):
...
+    def size(self) -> int:
+        """
+        Size of type
+        :return: int
+        """
+
+    def base_qtype(self):
+        return self._qtype.base_qtype()
+
+    def __eq__(self, other):
```
jbrezmorf: In which context we use the comparison?
Rather remove it. Two types of the same size are not equivalent.

src/mlmc/quantity_concept.py
```diff
@@ -121,120 +373,594 @@ def mult_op(x, y):
...
+        if isinstance(other, QType):
+            return self.size() == other.size()
+        return False
+
+    def replace_scalar(self, new_qtype):
```
jbrezmorf, line 790: ? private

src/mlmc/quantity_concept.py
```diff
@@ -121,120 +373,594 @@ def mult_op(x, y):
...
+    def _keep_dims(self, chunk):
+        """
+        Always keep chunk dimensions to be [M, chunk size, 2]
+        :param chunk: list
+        :return: list
+        """
+        # Keep dims [M, chunk size, 2]
+        if len(chunk.shape) == 2:
+            chunk = chunk[np.newaxis, :]
+        elif len(chunk.shape) > 2:
```
jbrezmorf, line 812: What about `len(chunk.shape) > 3`?

src/mlmc/quantity_concept.py
```diff
@@ -121,120 +373,594 @@ def mult_op(x, y):
...
+            self._qtype.replace_scalar(new_qtype)
+
+    def _keep_dims(self, chunk):
+        """
+        Always keep chunk dimensions to be [M, chunk size, 2]
+        :param chunk: list
+        :return: list
+        """
+        # Keep dims [M, chunk size, 2]
+        if len(chunk.shape) == 2:
```
jbrezmorf, line 810: I assume the shape (chunk_size, 2) here, I'm not sure if this can happen. 
If we want this method to be general we should also call `assert len(chunk.shape) > 1`.

src/mlmc/quantity_concept.py
jbrezmorf: E.g.
```
>>>  array([0, 1, 2])[1:2].shape
(1,)
```

src/mlmc/quantity_concept.py
```diff
@@ -121,120 +373,594 @@ def mult_op(x, y):
...
-        return getattr(key)
+class ArrayType(QType):
+    def __init__(self, shape, qtype: QType, start=0):
+        self._shape = shape
+        self._qtype = qtype
+        self.start = start
 
-    def max(self):
+    def size(self) -> int:
+        return np.prod(self._shape) * self._qtype.size()
```
jbrezmorf, line 849: Do we support operations for e.g.:
```
ArrayType(shape, 
    DictType([
        ('conc', ScalarType()), 
        ('vel', ArrayType((3,), ScalarType()))
        ]))
```
That can be done and possibly it can work but has to be carefully tested.
However, QuantityMean definitely needs an improvement in order to support that.

src/mlmc/quantity_concept.py
```diff
@@ -121,120 +373,594 @@ def mult_op(x, y):
...
+        """
+        # Keep dims [M, chunk size, 2]
+        if len(chunk.shape) == 2:
+            chunk = chunk[np.newaxis, :]
+        elif len(chunk.shape) > 2:
+            chunk = chunk.reshape((np.prod(chunk.shape[:-2]), chunk.shape[-2], chunk.shape[-1]))
+
+        return chunk
+
+    def _make_getitem_op(self, chunk, new_qtype, key=None):
```
jbrezmorf: Move this into DictType? ArrayType has its own implementation.

src/mlmc/quantity_concept.py
```diff
@@ -121,120 +373,594 @@ def mult_op(x, y):
...
+        """
+        ArrayType indexing
+        :param key: int, tuple of ints or slice objects
+        :return: QuantityType - ArrayType or self._qtype
+        """
+        # Get new shape
+        new_shape = np.empty(self._shape)[key].shape
+
+        # One selected item is considered to be a scalar QType
+        if len(new_shape) == 1 and new_shape[0] == 1:
```
jbrezmorf, line 861: Why this? It breaks mimicking numpy and its logic. E.g.
`array[1:2]` is not `array[1]`

src/mlmc/quantity_concept.py
```diff
@@ -121,120 +373,594 @@ def mult_op(x, y):
...
+            new_shape = ()
+
+        # Result is also array
+        if len(new_shape) > 0:
```
jbrezmorf, line 865: This is OK.

src/mlmc/quantity_concept.py
```diff
@@ -121,120 +373,594 @@ def mult_op(x, y):
...
+            q_type = ArrayType(new_shape, qtype=copy.deepcopy(self._qtype))
```
jbrezmorf: Why we need deep copies? Add a comment.
I wonder since QType should remain unchanged once constructed so no copy should be necessary.

src/mlmc/quantity_concept.py
```diff
@@ -121,120 +373,594 @@ def mult_op(x, y):
...
-    def __init__(self, *args):
-        assert( all( (a.is_array() for a in args) ) )
-        assert( all( (a.shape == args[0].shape for a in args) ) )
-        self.q_list = args
-        self.shape = [len(self.q_list)] + args[0].shape
+    def size(self) -> int:
+        return len(self._times) * self._qtype.size()
 
+    def __getitem__(self, key):
+        if key not in self._times:
```
jbrezmorf: Catch exception when calling `times.index`, single lookup, and thread save.

src/mlmc/quantity_concept.py
```diff
@@ -121,120 +373,594 @@ def mult_op(x, y):
...
+        """
+        self._dict = dict(args)
+        self._qtype = args[0][1]
+        self.start = start
+        assert all(q_type == self._qtype for _, q_type in args)
+
+    def size(self) -> int:
+        return len(self._dict.keys()) * self._qtype.size()
+
+    def __getitem__(self, key):
```
jbrezmorf, line 934: This should currently work in the same way as for the DictType. Can change later so we can keep it as a duplicate code.

src/mlmc/quantity_concept.py
```diff
@@ -121,120 +373,594 @@ def mult_op(x, y):
...
+    def __init__(self, args: List[Tuple[str, QType]]):
+        self._dict = dict(args)  # Be aware we it is ordered dictionary
+        self.start = 0
+
+        self._check_base_type()
+
+    def _check_base_type(self):
+        qtypes = list(self._dict.values())
+        for qtype in qtypes[1:]:
+            if not isinstance(qtype.base_qtype(), type(qtypes[0].base_qtype())):
```
jbrezmorf: introduce a variable for `qtypes[0].base_qtype()`

src/mlmc/quantity_concept.py
```diff
@@ -121,120 +373,594 @@ def mult_op(x, y):
...
+                raise TypeError("qtype {} has base QType {}, expecting {}. "
+                                "All QTypes must have same base QType, either SacalarType or BoolType".
+                                format(qtype, qtype.base_qtype(), qtypes[0].base_qtype()))
+
+    def base_qtype(self):
+        return list(self._dict.values())[0].base_qtype()
```
jbrezmorf, line 964: What about `next(iter(self._dict.values())).base_qtype()`?

src/mlmc/quantity_concept.py
```diff
@@ -121,120 +373,594 @@ def mult_op(x, y):
...
+
+    def size(self) -> int:
+        return int(np.sum(q_type.size() for _, q_type in self._dict.items()))
```
jbrezmorf, line 967: `for q_type in self.get_qtypes()` 

however that should be made to a property

src/mlmc/quantity_concept.py
```diff
@@ -121,120 +373,594 @@ def mult_op(x, y):
...
+
+    def replace_scalar(self, new_qtype):
+        for key, qtype in self._dict.items():
+            if isinstance(qtype, ScalarType):
+                self._dict[key] = new_qtype
+            else:
+                qtype.replace_scalar(new_qtype)
+
+    def __getitem__(self, key):
+        if key not in self._dict:
```
jbrezmorf: Just catch exception or possibly leave the default exception. Do not list all keys in the exception, possibly just a few of them.

jbrezmorf (commented):
Review completed.
Nothing critical, but various suggestions how to improve the code.

src/mlmc/quantity_concept.py
jbrezmorf: This is OK.

src/mlmc/quantity_concept.py
jbrezmorf: Make issue.

src/mlmc/quantity_concept.py
jbrezmorf: Fix it in general ufunc implementation and then we can remove this special check.

src/mlmc/quantity_concept.py
jbrezmorf: Resulting quantity would have selection_id of itself:

```
q = Quantity(...)
q._selection_id = id(q)
return q
```

src/mlmc/quantity_concept.py
jbrezmorf: Make a separate issue as this change can possibly break another code.

src/mlmc/quantity_concept.py
jbrezmorf: Possibly move both checks into _reuce_op and use it here.

src/mlmc/quantity_concept.py
jbrezmorf: Bootstrapping review in a separate issue.

src/mlmc/quantity_concept.py
jbrezmorf: Make private and introduce functions: 
- QArray(list of quantities or list of list of quantities ...) , like `np.array`
- QDict(list of pairs (key, quantity))
- QTimeSeries(list of pairs (time, quantity))
- QField( list of pairs (key, quantity))

src/mlmc/quantity_concept.py
jbrezmorf: Seems correct, just explain better what it returns.

src/mlmc/quantity_concept.py
jbrezmorf: Make a separate issue.

src/mlmc/quantity_concept.py
martinspetlik: add assert

src/mlmc/quantity_concept.py
jbrezmorf: Possibly wrong use of deprecated start and end attributes.

src/mlmc/quantity_concept.py
martinspetlik: It doesn't work with quantities but with numbers.

src/mlmc/quantity_concept.py
martinspetlik: I don't think so.

src/mlmc/quantity_concept.py
martinspetlik: I think checking `base_qtype` in ufunc implementation isn't a good solution. We might have discussed a situation when different function arguments can be of a different types. So I cannot think of a general solution for that.

src/mlmc/quantity_concept.py
martinspetlik: I think `_reduction_op` is not suitable for this. We should discuss it.

src/mlmc/quantity_concept.py
martinspetlik: Shape (chunk_size, 2) can occur when we estimating the mean of a scalar quantity.

src/mlmc/quantity_concept.py
martinspetlik: `len(chunk.shape) > 3` will be handled by `elif`

src/mlmc/quantity_concept.py
martinspetlik: We use `qtype.start` for selection, e.g. `__getitem__` . 
I sometimes want to get the current qtype for some further use but  I don't want to use the same `start` index so I use the deep copy.

src/mlmc/quantity_concept.py
martinspetlik: QArray should receive shape...

src/mlmc/quantity_concept.py
martinspetlik: I don't think so.

src/mlmc/quantity_concept.py
martinspetlik: separate issue

src/mlmc/quantity_concept.py
martinspetlik: No usage -> removed

src/mlmc/quantity_concept.py
martinspetlik: Separate issue, it is connected to `QuantityMean__getitem__` bugs.

src/mlmc/quantity_concept.py
martinspetlik: It  keeps chunk shape to be [M, chunk size, 2]. Sometimes we need to 'flatten' first few shape to have desired chunk shape.
For scalar quantities, the input block can have the shape (chunk size, 2)

src/mlmc/quantity_concept.py
martinspetlik: Separate issue, it is connected to QuantityMean__getitem__ bugs

src/mlmc/quantity_concept.py
martinspetlik: It is used by DictType and FieldType.

src/mlmc/quantity_concept.py
jbrezmorf: OK. ponechat.

src/mlmc/quantity_concept.py
jbrezmorf: OK

src/mlmc/quantity_concept.py
jbrezmorf: OK

src/mlmc/quantity_concept.py
jbrezmorf: OK

src/mlmc/quantity_concept.py
jbrezmorf: Agree.

src/mlmc/quantity_concept.py
jbrezmorf: Wrap is applied in `_mask_quantity`.

src/mlmc/quantity_concept.py
jbrezmorf: Samples should always work with shape `(M,N,2)`, possibly with M=1.

src/mlmc/quantity_concept.py
jbrezmorf: OK, but since that should not happen, it should be rather `assert`.

src/mlmc/quantity_concept.py
jbrezmorf: Input `chunk` should have shape `(M,N,2)` and after slicing we change only the value of `M`.

src/mlmc/quantity_concept.py
jbrezmorf: Turn get item operation into slicing operation and drop the start attribute of the QType. Then we do not need the deep copy.

src/mlmc/quantity_concept.py
jbrezmorf: OK. Issue is e.g. `chunk[1]` which converts np.array to float and QArray into QScalar.

PR #127, opened 2020-12-11
Ms successful dir

martinspetlik:


src/mlmc/sampling_pool.py
```diff
@@ -158,33 +157,22 @@ def handle_sim_files(work_dir, sample_id, level_sim):
...
         :return: None
         """
         if sample_workspace and work_dir is not None:
-            failed_dir = SamplingPool._create_failed(work_dir)
-            sample_dir = SamplingPool.change_to_sample_directory(work_dir, sample_id)
-            if os.path.exists(os.path.join(failed_dir, sample_id)):
-                shutil.rmtree(os.path.join(failed_dir, sample_id), ignore_errors=True)
-            shutil.copytree(sample_dir, os.path.join(failed_dir, sample_id))
-            shutil.rmtree(sample_dir, ignore_errors=True)
+            if int(sample_id[-1:]) < 5:
```
jbrezmorf, line 170: What is meaning of this test? What is special about number 5?  Seems that it has changed to allow moves not only for failed samples, but also for a few correct samples. If the method makes the move, then the logic if some move should be performed must be out of it.

jbrezmorf (commented):
Just one point to discuss.

src/mlmc/sampling_pool.py
martinspetlik: You are right, the condition must definitely be outside of this method. This is condition for moving several (5) successfulI samples, so not suitable for moving failed samples. I created separate issue for it.

PR #128, opened 2020-12-12
Ms quantity estimator

martinspetlik:


src/mlmc/tool/flow_mc.py
```diff
@@ -31,15 +30,15 @@ def create_corr_field(model='gauss', corr_length=0.125, dim=2, log=True):
         ])
 
     if model == 'exp':
-        model = gstools.Exponential(dim=dim, len_scale=len_scale)
+        model = gstools.Exponential(dim=dim, var=sigma**2, len_scale=len_scale)
```
martinspetlik, line 33: This is obsolete, `var` shouldn't be set in covariance model in our implementation.

src/mlmc/quantity_estimate.py
```diff
@@ -1,312 +0,0 @@
-import numpy as np
```
martinspetlik, line 1: This file was replaced by `mlmc.estimator`
It has nothing common with future `quantity_estimate` file.

src/mlmc/tool/pbs_job.py
```diff
@@ -173,7 +173,7 @@ def calculate_samples(self):
                 success.append((current_level, sample_id, (res[0], res[1])))
                 # Increment number of successful samples for measured time
                 SamplingPool.move_dir(sample_id, level_sim.need_sample_workspace, self._output_dir,
-                                      dest_dir=SamplingPool.SEVERAL_SUCCESSFUL_DIR)
+                                      dest_dir=os.path.join(self._output_dir, SamplingPool.SEVERAL_SUCCESSFUL_DIR))
```
jbrezmorf: Usea alocal variable.

test/01_cond_field/process_simple.py
```diff
@@ -64,11 +64,10 @@ def __init__(self):
 
     def process(self):
         sample_storage = SampleStorageHDF(file_path=os.path.join(self.work_dir, "mlmc_{}.hdf5".format(self.n_levels)))
-        sample_storage.chunk_size = 1024
+        sample_storage.chunk_size = 1e8
```
jbrezmorf, line 67: We should test with smaller chunk size to test the processing via multiple chunks.

test/01_cond_field/process_simple.py
```diff
@@ -98,29 +95,33 @@ def process(self):
...
 
     def construct_density(self, estimator, tol=1.95, reg_param=0.0):
         """
         Construct approximation of the density using given moment functions.
-        Args:
-            moments_fn: Moments object, determines also domain and n_moments.
-            tol: Tolerance of the fitting problem, with account for variances in moments.
-                 Default value 1.95 corresponds to the two tail confidency 0.95.
-            reg_param: Regularization parameter.
+        :param quantity: mlmc.quanitity.Quantity instance, quantity for which the density is reconstructed
```
jbrezmorf: There are no parameters `quantity`, `moments_fn`, and `sample_storage` !?

test/01_cond_field/process_simple.py
```diff
@@ -98,29 +95,33 @@ def process(self):
...
+        :param tol: Tolerance of the fitting problem, with account for variances in moments.
+                    Default value 1.95 corresponds to the two tail confidence 0.95.
+        :param reg_param: regularization parameter
+        :return: None
         """
         distr_obj, result, _, _ = estimator.construct_density(tol=tol, reg_param=reg_param)
-        #distr_plot = mlmc.tool.plot.Distribution(title="{} levels, {} moments".format(self.n_levels, self.n_moments))
-        distr_plot = mlmc.tool.plot.ArticleDistribution(title="{} levels, {} moments".format(self.n_levels, self.n_moments))
+        # distr_plot = mlmc.tool.plot.Distribution(title="{} levels, {} moments".format(self.n_levels, self.n_moments))
+        distr_plot = mlmc.tool.plot.ArticleDistribution(
```
jbrezmorf, line 121: Article specific stuff should rather be placed into a tagged branch out of the master. 
1. It allows various hacks without disturbing the master.
2. It keeps specific changes related to the reason (particular publication).

test/01_cond_field/process_simple.py
```diff
@@ -319,14 +320,15 @@ def all_collect(self, sampler):
...
         Calculate moments through the mlmc.QuantityEstimate
         :param sampler: mlmc.Sampler
         :return: None
         """
         # Simple moment evaluation
-        moments_fn = self.set_moments(sampler.sample_storage)
+        moments_fn = self.set_moments(sampler._sample_storage)
 
-        q_estimator = QuantityEstimate(sample_storage=sampler.sample_storage, moments_fn=moments_fn,
+        q_estimator = QuantityEstimate(sample_storage=sampler._sample_storage, moments_fn=moments_fn,
```
jbrezmorf, line 329: However `quantity_estimate.py` is deleted in this very PR.
And the import is removed.
We have to manually check and fix this real test after all changes.

src/mlmc/tool/pbs_job.py
martinspetlik: Local variable for `os.path.join(self._output_dir, SamplingPool.SEVERAL_SUCCESSFUL_DIR)`?

test/01_cond_field/process_simple.py
martinspetlik: It is not effective to have a small chunk in this place, it takes a long time.

test/01_cond_field/process_simple.py
martinspetlik: OK, but I would do it manually at the end, because there are just few compact methods and class.

test/01_cond_field/process_simple.py
martinspetlik: I think `calculate_method` is no longer in use in the end. Either way I use this file a lot and eventually, everything works.

test/01_cond_field/process_simple.py
jbrezmorf: Could take longer when usin very small chunks, but size about 1e5 should make no difference. 
Nevertheless the point was, that this may serve as a real test of the chunk-based estimate algorithm.

PR #129, opened 2020-12-12
Ms violinplot

martinspetlik:


src/mlmc/estimator.py
```diff
@@ -301,6 +301,27 @@ def plot_var_compare(self, nl):
     def plot_var_var(self, nl):
         self[nl].plot_bootstrap_var_var(self._moments_fn)
 
+    def fine_coarse_violinplot(self):
```
jbrezmorf, line 304: Pandas is a nontrivial dependency that can make compatibility issues. Possibly move the creation of the frame from the sample storage into `tools/plot.py` as well.

src/mlmc/tool/plot.py
```diff
@@ -1,6 +1,33 @@
...
+#     ],
+# }
+# mpl.rcParams.update(pgf_with_pdflatex)
+
+
+
+# mpl.rcParams['xtick.labelsize']=12
+# mpl.rcParams['ytick.labelsize']=12
+
+import matplotlib
```
jbrezmorf, line 5: So, remove `import matplotlib as mpl`.

src/mlmc/tool/plot.py
```diff
@@ -1042,3 +1457,101 @@ def plot_error(arr, ax, label):
...
+    ax = fig.add_subplot(1, 1, 1)
+    ax.set_xscale('log')
+    lbls = ['{}'.format(nl) for nl in n_elements]
+    ax.set_xticklabels(lbls)
+    #ax.set_yscale('log')
+    ax.plot(1/(level_params**2), n_ops)
+    _show_and_save(fig, "flow_time", "flow_time")
+
+
+if __name__ == "__main__":
```
jbrezmorf: ??

jbrezmorf: I suggest making a branch for the preparation of the paper. And then remove (or comment) 
plots that have no general usage.

src/mlmc/estimator.py
martinspetlik: sample storage doesn't create data frame, it is done here in estimator, so I think there is no use in moving it from estimator to tools/plot.py

jbrezmorf: The data frame is created only for plotting so it is part of plotting and thus should be in the `tools/plot.py`.

PR #130, opened 2020-12-12
Ms quantity dividing

martinspetlik:


src/mlmc/estimator.py
```diff
@@ -2,7 +2,7 @@
 import scipy.stats as st
 import scipy.integrate as integrate
 from mlmc.tool import plot
-from mlmc.quantity_estimate import estimate_mean, moments, covariance
+import mlmc.quantity_estimate as qe
```
jbrezmorf, line 5: I.e. that is better, I prefer this kind of import as it prevents possible name clash.

src/mlmc/estimator.py
```diff
@@ -351,12 +351,11 @@ def construct_density(self, tol=1e-8, reg_param=0.0, orth_moments_tol=1e-4, exac
         """
         Construct approximation of the density using given moment functions.
         """
-        cov = estimate_mean(covariance(self._quantity, self._moments_fn))()
+        cov = np.squeeze(qe.estimate_mean(qe.covariance(self._quantity, self._moments_fn))())
```
jbrezmorf: Introduce a local variable:
```
cov_mean = qe.estimate_mean(qe.covariance(self._quantity, self._moments_fn))
cov_mat = cov_mean()
```
That should already produce the covariance matrix with shape $(R,R)$ where $R$ is the number of moments.

That why is the `squeeze` necessary. I find it potentialy dangerous as it can reduce number of axes and break consistency of the data structure. It seems to me that it can be usefull for scripting but should rather not to be used in a library.

src/mlmc/quantity.py
```diff
@@ -168,6 +153,8 @@ def _reduction_op(self, quantities, operation):
         for quantity in quantities:
             if not isinstance(quantity, QuantityConst):
                 return Quantity(quantity.qtype, operation=operation, input_quantities=quantities)
+        # Quantity from QuantityConst instances
+        return QuantityConst(quantities[0].qtype, value=operation(*[q._value for q in quantities]))
```
jbrezmorf, line 155: Can we move this kind of optimization to the more general level? I imagine that we can replace calls to Quantity constutctor by a static method that returns eithr Quantity of QuantityConst as is done here.

src/mlmc/quantity.py
```diff
@@ -168,6 +153,8 @@ def _reduction_op(self, quantities, operation):
         for quantity in quantities:
```
jbrezmorf, line 151: Can use more readable variant:
```
is_const_quantity = (isinstance(q, QuantityConst) for q in quantities)
if any(is_const_quantity):
    return ...
```

src/mlmc/quantity.py
```diff
@@ -561,13 +549,14 @@ def __init__(self, quantity_type, mean, var, l_means=[], l_vars=[], n_samples=No
         self._l_means = np.array(l_means)
         self._l_vars = np.array(l_vars)
         self._n_samples = n_samples
+        self._n_rm_samples = n_rm_samples
 
     def __call__(self):
```
jbrezmorf, line 571: Rather keep only the `mean`property, that is more descriptive and there should be preferably only one way to write it.

src/mlmc/quantity.py
```diff
@@ -590,13 +579,17 @@ def n_samples(self):
...
 
     def _reshape(self, data, levels=False):
-        # if isinstance(self.qtype, ArrayType):
-        #     reshape_shape = self.qtype._shape
-        #     if isinstance(reshape_shape, int):
-        #         reshape_shape = [reshape_shape]
-        #     if levels:
-        #         return data.reshape((data.shape[0], *reshape_shape))
-        #     return data.reshape(*reshape_shape)
+        if isinstance(self.qtype, qt.ArrayType):
```
jbrezmorf, line 599: Define what should be purpose of this method, currently it just performs several different things which can possibly be done easier at calling side.

src/mlmc/quantity.py
```diff
@@ -590,13 +579,17 @@ def n_samples(self):
...
+            reshape_shape = self.qtype._shape
+            size = self.qtype._qtype.size()
+            if isinstance(reshape_shape, int):
```
jbrezmorf, line 602: If the QuantityMean is in the consisten state this should not happen.

src/mlmc/quantity.py
```diff
@@ -590,13 +579,17 @@ def n_samples(self):
...
+                reshape_shape = [reshape_shape]
+            if levels:
```
jbrezmorf, line 604: levels=True should be treated at calling side, e.g. using a comprehension

src/mlmc/quantity.py
```diff
@@ -590,13 +579,17 @@ def n_samples(self):
...
+                return data.reshape((data.shape[0], *reshape_shape))
+            if size > 1:
```
jbrezmorf, line 606: How this happens?

src/mlmc/quantity.py
```diff
@@ -630,25 +618,19 @@ def __getitem__(self, key):
         l_vars = self._l_vars
 
         if reshape_shape is not None:
-            if newshape is not None:  # reshape [Mr] to e.g. [..., R, R, M]
-                mean = mean.reshape(*reshape_shape, *newshape)
-                var = var.reshape(*reshape_shape, *newshape)
-                l_means = l_means.reshape((l_means.shape[0], *reshape_shape, *newshape))
-                l_vars = l_vars.reshape((l_vars.shape[0], *reshape_shape, *newshape))
-            elif (np.prod(mean.shape) // np.prod(reshape_shape)) > 1:
+            if (np.prod(mean.shape) // np.prod(reshape_shape)) > 1:
```
jbrezmorf, line 638: Isn't mean.shape always single axis? Then `mean.shape[0]` would be more specific.

src/mlmc/quantity.py
jbrezmorf: Postponed to #145

src/mlmc/quantity.py
jbrezmorf: Postponed to #146

jbrezmorf (commented):
Most of the comments already covered by separate issues. 
Just deal with those without denoted relation to any issue.

src/mlmc/estimator.py
martinspetlik: Presumably, `np.squeeze` is no longer necessary.

src/mlmc/quantity.py
martinspetlik: It is more readable, but it requires another `for` to find first `Quantity`instance in case of `if any(is_const_quantity):` is `False`.

src/mlmc/quantity.py
martinspetlik: It was originally required to support `__call__`. I think it is useful and I use it in many places. I don't see any disadvantages of support for both methods.

On the other hand, there are other issues in `QuantityMean`, so removing `__call__` wouldn't be a big deal for me.

src/mlmc/quantity.py
martinspetlik: If `size == 1`  then `self.qtype._qtype` is scalar, so no `reshape` is possible

src/mlmc/quantity.py
martinspetlik: comment to #145 added

src/mlmc/quantity.py
martinspetlik: comment to #145 added

src/mlmc/quantity.py
martinspetlik: #155

src/mlmc/quantity.py
jbrezmorf: The point of using `any` is to avoid the loop:
```
>>> l = (x>3 for x in range(10))
>>> any(l)
True
```

src/mlmc/quantity.py
jbrezmorf: I know, that was the first proposal to have it close to the situation before QuantityMean was introduced. But now I see it problematic QuantityMan does not have a clear functor semantics and `qmean.mean` is clear to a newcomer but `gmean()` is not.

src/mlmc/quantity.py
jbrezmorf: Possible, but not necessary once `reshape_shape == []`:
```
>>> x = np.array([1])
>>> x.reshape([])
array(1)
```

src/mlmc/quantity.py
martinspetlik: comment added to #146

PR #132, opened 2020-12-14
Ms quantity improv

martinspetlik:


src/mlmc/moments.py
```diff
@@ -102,18 +102,12 @@ def _center(self, value):
             return value - self.mean
 
         if not isinstance(self.mean, int):
-            if np.all(value[..., 1]) == 0:
-                value[..., 0] = value[..., 0] - self.mean[:, None]
-            else:
-                value[...] = value[...] - self.mean[:, None, None]
+            value[...] = value[...] - self.mean[:, None, None]
```
martinspetlik, line 105: This PR contains change of shape for level 0

src/mlmc/estimator.py
```diff
@@ -106,8 +107,7 @@ def estimate_diff_vars_regression(self, n_created_samples, moments_fn=None, raw_
             if moments_fn is None:
                 moments_fn = self._moments_fn
             raw_vars, n_samples = self.estimate_diff_vars(moments_fn)
-        sim_steps = self._sample_storage.get_level_parameters()
-
+        sim_steps = np.squeeze(self._sample_storage.get_level_parameters())
```
jbrezmorf, line 110: Why using `np.squeeze`? Try to avoid it as it can create inconistant array shape, by reducing axes of the size one. 

Here `sim_steps` should be a vector. Can be wrong for the single level MC.

src/mlmc/moments.py
jbrezmorf: The shape should change, but not the number of axes.  Nevertheless this code is more specific.

src/mlmc/quantity.py
```diff
@@ -64,6 +64,20 @@ def get_quantity_storage(self):
...
+        Set selection id
+        selection id is None by default,
+         but if we create new quantity from quantities that are result of selection we need to pass selection id
+        """
+        selection_id = None
+        for input_quantity in self._input_quantities:
+            if selection_id is None:
+                selection_id = input_quantity.selection_id()
+            elif input_quantity.selection_id() is not None and selection_id != input_quantity.selection_id():
+                return None
```
jbrezmorf: This case should rise and exception.

src/mlmc/quantity.py
```diff
@@ -86,18 +100,16 @@ def _op_additional_params(self):
         """
         self._additional_params = {}
         sig_params = signature(self._operation).parameters
-        if 'level_id' in sig_params:
-            self._additional_params['level_id'] = 0
-        if 'i_chunk' in sig_params:
-            self._additional_params['i_chunk'] = 0
+        if 'chunk_spec' in sig_params:
+            self._additional_params['chunk_spec'] = None
```
jbrezmorf, line 104: We have to get rid of `additional_params`, seems that the only remaining place is the `subsample`.

src/mlmc/quantity_estimate.py
```diff
@@ -49,21 +48,20 @@ def estimate_mean(quantity, chunk_size=512000000, level_means=False):
...
                     if chunk_id == 0:
                         sums = [np.zeros(chunk.shape[0]) for _ in range(n_levels)]
                         sums_of_squares = [np.zeros(chunk.shape[0]) for _ in range(n_levels)]
-                    # Coarse result for level 0, there is issue for moments_fn processing (not know about level)
-                    chunk[..., 1] = 0
-
-                chunk, n_rm_samples = mask_nan_samples(chunk)
+                    chunk_diff = chunk[:, :, 0]
+                else:
+                    chunk_diff = chunk[:, :, 0] - chunk[:, :, 1]
```
jbrezmorf, line 63: There was some particular motivation for changing the shape of the Level 0. Seems that it leads to few other complications in the code, but I do not see the simplification in this PR.

src/mlmc/quantity_spec.py
```diff
@@ -21,3 +21,12 @@ class ChunkSpec:
     # Number of samples which we want to retrieve
     chunk_size: int = 512000000
     # Chunk size in bytes in decimal, determines number of samples in chunk
+
+    def __hash__(self):
+        return hash((self.level_id, self.chunk_id, self.n_samples, self.chunk_size))
```
jbrezmorf, line 26: Why this custom implementation it seems to be equivalent to the default implementation provided by attr.s

src/mlmc/sample_storage.py
```diff
@@ -240,7 +240,16 @@ def sample_pairs_level(self, chunk_spec):
             results = self._results[int(chunk_spec.level_id)]
```
jbrezmorf: Rather set `n_samples` if they are not provided by the chunk_spec. No need to have two different code branches.

src/mlmc/tool/hdf5.py
```diff
@@ -379,30 +379,29 @@ def collected(self, chunk_spec=None):
...
         """
-        if self._collected_n_items is None:
-            with h5py.File(self.file_name, 'r') as hdf_file:
-                if 'collected_values' not in hdf_file[self.level_group_path]:
-                    return None
-                dataset = hdf_file["/".join([self.level_group_path, "collected_values"])]
-                self._collected_n_items = len(dataset[()])
-        return self._collected_n_items
+        with h5py.File(self.file_name, 'r') as hdf_file:
+            if 'collected_values' not in hdf_file[self.level_group_path]:
```
jbrezmorf, line 400: Should we rise an exception?

test/01_cond_field/process_simple.py
```diff
@@ -217,7 +217,7 @@ def set_environment_variables(self):
             self.init_sample_timeout = 60
             self.sample_timeout = 60
             self.flow123d = "/home/jb/workspace/flow123d/bin/fterm flow123d dbg"
```
jbrezmorf, line 219: We should have a sort of setting for the complex tests, see #154

src/mlmc/estimator.py
martinspetlik: It works well in the current implementation. 

`level_parameters`: `[[step_1], [step_2], [step_3], ...]`

src/mlmc/quantity.py
martinspetlik: #156

src/mlmc/quantity_spec.py
martinspetlik: I got `TypeError: unhashable type: 'ChunkSpec'`. I have no idea why it doesn't work by default.

src/mlmc/quantity_estimate.py
martinspetlik: Simplifications:
- param `level_id` is no longer used in `Quantity._process_mask`
- auxiliary zeros at level one caused difficulties in `mask_nan_samples` for `__truediv__` or `__mod__`. Currently it works fine without additional code.

jbrezmorf:
See attr.s doc on hashing: https://www.attrs.org/en/stable/hashing.html
We should use `@attr.s(frozen=True)`.

PR #137, opened 2020-12-16
Ms estimate tests fix

martinspetlik:


martinspetlik:
Some test errors are solved in later commits. So fix would be waste of time.

src/mlmc/quantity.py
```diff
@@ -7,7 +7,9 @@
 from scipy import interpolate
 from typing import List, Tuple
 from mlmc.sample_storage import SampleStorage
-from mlmc.sim.simulation import QuantitySpec
+from mlmc.quantity_spec import QuantitySpec
+from mlmc.quantity_types import QType, ScalarType, BoolType, ArrayType, DictType, TimeSeriesType, FieldType
```
martinspetlik, line 11: no quantity_types file in this phase

src/mlmc/quantity.py
```diff
@@ -154,8 +168,7 @@ def eval_moments(x):
     # Create quantity type which has moments at the bottom
     if mom_at_bottom:
         moments_array_type = ArrayType(shape=(moments_fn.size,), qtype=ScalarType())
-        moments_qtype = copy.deepcopy(quantity.qtype)
-        moments_qtype.replace_scalar(moments_array_type)
+        moments_qtype = QType.replace_scalar(copy.deepcopy(quantity.qtype), moments_array_type)
```
martinspetlik, line 170: Other commits deal with deepcopy

src/mlmc/quantity.py
```diff
@@ -487,6 +501,30 @@ def mask_gen(x, *args):
             return mask
         return self._mask_quantity(size, mask_gen)
 
+    def subsample(self, sample_vec):
```
martinspetlik, line 503: key bootstrap function
I think there will be changes such as ChunkSpec, ...

test/test_quantity_concept.py
```diff
@@ -627,6 +644,64 @@ def test_moments(self):
         value_mean = location_mean[0]
         assert len(value_mean()) == 1
 
+    @pytest.mark.parametrize("memory", [False, True])
```
martinspetlik, line 647: join with test_moments at the very end

src/mlmc/quantity.py
```diff
@@ -680,19 +717,46 @@ def __init__(self, quantity_type, mean, var):
...
+
+    @property
+    def l_vars(self):
+        return self._reshape(self._l_vars, levels=True)
+
+    @property
+    def n_samples(self):
+        return self._n_samples
+
+    def _reshape(self, data, levels=False):
```
martinspetlik, line 751: This function causes some tests fail, there are improvents in later commits.

src/mlmc/quantity.py
```diff
@@ -115,12 +116,24 @@ def estimate_mean(quantity):
 
     mean = np.zeros_like(sums[0])
     var = np.zeros_like(sums[0])
+    l_means = []
```
jbrezmorf, line 119: Good right now. I propose to pass level means and vars always to the QuantityMean, see #143.

src/mlmc/quantity.py
jbrezmorf: OK

src/mlmc/quantity.py
```diff
@@ -186,9 +198,8 @@ def eval_cov(x):
 
     # Create quantity type which has covariance matrices at the bottom
     if cov_at_bottom:
-        moments_array_type = ArrayType(shape=(moments_fn.size, moments_fn.size,), qtype=ScalarType())
-        moments_qtype = copy.deepcopy(quantity.qtype)
-        moments_qtype.replace_scalar(moments_array_type)
+        moments_array_type = ArrayType(shape=(moments_fn.size, moments_fn.size, ), qtype=ScalarType())
+        moments_qtype = QType.replace_scalar(copy.deepcopy(quantity.qtype), moments_array_type)
```
jbrezmorf, line 202: It seems that `replace_scalar` should be a proper method, so call simplifies to:
`the_deep_copy.replace_scalar(moments_arry_type)`, returning the new instance of the QType so performing a copy anyway.

src/mlmc/quantity.py
jbrezmorf: #144

src/mlmc/quantity.py
jbrezmorf: What is `data`? Rename to something more specific and provide a comment or type hint.
What is the meaning of the `levels` flag? It seems to me that there should possibly be two methods, one calling the other.

src/mlmc/quantity.py
```diff
@@ -719,21 +785,48 @@ def __getitem__(self, key):
...
         mean = self._mean
         var = self._var
+        l_means = self._l_means
+        l_vars = self._l_vars
 
         if reshape_shape is not None:
             if newshape is not None:  # reshape [Mr] to e.g. [..., R, R, M]
-                mean = mean.reshape((*reshape_shape, *newshape))
-                var = var.reshape((*reshape_shape, *newshape))
+                mean = mean.reshape(*reshape_shape, *newshape)
```
jbrezmorf, line 793: Try to simplify this method:
1. do not store mean and var, only l_mean and l_var
2. reduce number of special cases, keep QuantityMean in a single consistent structure. e.g. neither reshape_shape nor newshape should be None (fill them with proper values before). 
3. If you find the special cases inevitable put there a comment to explain what are the individual cases.

jbrezmorf: Possibly also try to break the method into two parts:
1. convert general `key` argument to a particular slice ... via QType
2. extract part of l_means and l_vars (there should be a method for that to make the same extraction on both) ... independent of QType
...

src/mlmc/quantity.py
```diff
@@ -788,16 +881,35 @@ def size(self) -> int:
     def base_qtype(self):
         return self._qtype.base_qtype()
 
-    def replace_scalar(self, new_qtype):
+    @staticmethod
+    def replace_scalar(original_qtype, substitute_qtype):
```
jbrezmorf, line 885: Convert to proper method of QType.
This can be much simpler if implemented by recursive construction of the replaced type. Returning the new instance. Use separate implementation for dict and scalar instead of
`isinstance`.

src/mlmc/quantity.py
```diff
@@ -967,12 +1079,15 @@ def base_qtype(self):
     def size(self) -> int:
         return int(np.sum(q_type.size() for _, q_type in self._dict.items()))
 
-    def replace_scalar(self, new_qtype):
-        for key, qtype in self._dict.items():
+    @staticmethod
```
jbrezmorf, line 1082: How is this related to QType.replace_scalar? It is not even a classmethod. It seems to be already in use, but introduction of the probably same static method in QType is confusing.
Should be probably merged to a single method probably with virtual implementation in some QType subclases.

src/mlmc/tool/hdf5.py
```diff
@@ -157,6 +147,8 @@ def save_result_format(self, result_format, res_dtype):
                     dtype=result_format_dtype,
                     maxshape=(None,),
                     chunks=True)
+            else:
+                warnings.warn('Be careful, you are setting the new result format for an existing sample storage')
```
jbrezmorf: That should possibly be an error.

jbrezmorf (commented):
Major concerns about QuantityMean.__getitem__ implementation and QType.replace_scalar.
Few minor comments.

src/mlmc/quantity.py
martinspetlik: I think your suggestion is the same as previous version (`moments_qtype.replace_scalar(moments_array_type)`) which is replaced by this staticmethod call.
The main reason of this new implementation is that it allows you to replace scalar quantity by array of scalars which wasn't possible in the previous version.

src/mlmc/quantity.py
martinspetlik: There are significant changes in following commits, so issue created.

src/mlmc/quantity.py
martinspetlik: I think convert to proper method of QType is not good idea. I agree with separate implementation for dict and scalar.

I created new issue #147. I should think it over again. And it would be much easier with the newest version of code.

src/mlmc/quantity.py
martinspetlik: issue

PR #159, opened 2020-12-31
QuantityMean

### LINKED ISSUE #143: level means estimates
Always pass level means and vars to the MeanQuantity, possibly compute the total mean and var on request. That means a richer API of MeanQuantity.

### LINKED ISSUE #145: QuantityMean._reshape
`QuantityMean._reshape()`

What is `data`? Rename to something more specific and provide a comment or type hint.
What is the meaning of the`levels` flag? It seems to me that there should possibly be two methods, one calling the other.

### LINKED ISSUE #146: QuantityMean.__getitem__
`QuantityMean.__getitem__`

Try to simplify this method:
1.  do not store mean and var, only l_mean and l_var
2. reduce number of special cases, keep QuantityMean in a single consistent structure. e.g. neither reshape_shape nor newshape should be None (fill them with proper values before).
3. If you find the special cases inevitable put there a comment to explain what are the individual cases.

Possibly also try to break the method into two parts:

    

1. convert general key argument to a particular slice ... via QType
2. extract part of l_means and l_vars (there should be a method for that to make the same extraction on both) ... independent of QType

martinspetlik:
resolve #143
resolve #145
resolve #146

### DIFF
--- a/src/mlmc/quantity.py
+++ b/src/mlmc/quantity.py
@@ -538,126 +538,113 @@ class QuantityConst(Quantity):
 
 class QuantityMean:
 
-    def __init__(self, quantity_type, mean, var, l_means=[], l_vars=[], n_samples=None, n_rm_samples=0):
+    def __init__(self, quantity_type, l_means, l_vars, n_samples=None, n_rm_samples=None):
         """
-        QuantityMean represents result of mlmc.quantity_estimate.estimate_mean method
+        QuantityMean represents results of mlmc.quantity_estimate.estimate_mean method
         :param quantity_type: QType
-        :param mean: np.ndarray
-        :param var: np.ndarray
         :param l_means: np.ndarray, shape: L, M
         :param l_vars: np.ndarray, shape: L, M
-        :param n_samples: int, number of samples that were used for means computing
-        :param n_rm_samples: int, number of removed samples in mlmc.quantity_estimate.estimate_mean()
+        :param n_samples: List, number of samples that were used for means at each level
+        :param n_rm_samples: List, number of removed samples at each level,
+                             n_samples + n_rm_samples = all successfully collected samples
         """
         self.qtype = quantity_type
-        self._mean = mean
-        self._var = var
+        self._mean = None
+        self._var = None
         self._l_means = np.array(l_means)
         self._l_vars = np.array(l_vars)
-        self._n_samples = n_samples
-        self._n_rm_samples = n_rm_samples
+        self._n_samples = np.array(n_samples)
+        self._n_rm_samples = np.array(n_rm_samples)
 
-    def __call__(self):
+    def _calculate_mean_var(self):
         """
-        Return reshaped mean
+        Calculates the overall estimates of the mean and the variance from the means and variances at each level
         """
-        return self.mean
+        self._mean = np.sum(self._l_means, axis=0)
+        self._var = np.sum(self._l_vars / self._n_samples[:, None], axis=0)
+
+        # Sums full of zeros leads to NaN
+        if np.isnan(self._mean).all():
+            self._mean = []
+            self._var = []
 
     @property
     def mean(self):
+        if self._mean is None:
+            self._calculate_mean_var()
         return self._reshape(self._mean)
 
     @property
     def var(self):
+        if self._var is None:
+            self._calculate_mean_var()
         return self._reshape(self._var)
 
     @property
     def l_means(self):
-        return self._reshape(self._l_means, levels=True)
+        return np.array([self._reshape(means) for means in self._l_means])
 
     @property
     def l_vars(self):
-        return self._reshape(self._l_vars, levels=True)
+        return np.array([self._reshape(vars) for vars in self._l_vars])
 
     @property
     def n_samples(self):
         return self._n_samples
 
-    def _reshape(self, data, levels=False):
-        if isinstance(self.qtype, qt.ArrayType):
-            reshape_shape = self.qtype._shape
-            size = self.qtype._qtype.size()
-            if isinstance(reshape_shape, int):
-                reshape_shape = [reshape_shape]
-            if levels:
-                return data.reshape((data.shape[0], *reshape_shape))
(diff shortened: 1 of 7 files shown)


martinspetlik:
resolve #143 
resolve #145 
resolve #146

src/mlmc/quantity.py
```diff
@@ -538,126 +538,113 @@ def samples(self, chunk_spec):
...
         """
-        Return reshaped mean
+        Calculates the overall estimates of the mean and the variance from the means and variances at each level
         """
-        return self.mean
+        self._mean = np.sum(self._l_means, axis=0)
+        self._var = np.sum(self._l_vars / self._n_samples[:, None], axis=0)
+
+        # Sums full of zeros leads to NaN
+        if np.isnan(self._mean).all():
```
jbrezmorf, line 567: Why this special case?

src/mlmc/quantity.py
```diff
@@ -538,126 +538,113 @@ def samples(self, chunk_spec):
 
 class QuantityMean:
 
-    def __init__(self, quantity_type, mean, var, l_means=[], l_vars=[], n_samples=None, n_rm_samples=0):
+    def __init__(self, quantity_type, l_means, l_vars, n_samples=None, n_rm_samples=None):
```
jbrezmorf: Make n_samples and n_rm_samples obligatory. None values are not valid.

src/mlmc/quantity.py
jbrezmorf: Rather split this method into mean and var property functions.

src/mlmc/quantity.py
```diff
@@ -538,126 +538,113 @@ def samples(self, chunk_spec):
...
         """
-        Get items from Quantity, quantity type must support brackets access
+        Get item from current QuantityMean, quantity type must support brackets access
+        All levels means and vars are reshaped to their QType shape and then the item is gotten,
+        ath the end, new QuantityMean instance is created with flatten selected means and vars
         :param key: str, int, tuple
-        :return: np.ndarray
+        :return: QuantityMean
         """
         new_qtype, start = self.qtype[key]  # New quantity type
```
jbrezmorf: Seems, that the current behavior of qtype[key] violates the natural semantics of the indexing as it returns 'start' as well. 
In such a case we should rather use a regular method.

src/mlmc/quantity.py
```diff
@@ -538,126 +538,113 @@ def samples(self, chunk_spec):
...
+            new_shape = l_means.shape[0], *reshape_shape, np.prod(l_means.shape[1:]) // np.prod(reshape_shape)
+            l_means = l_means.reshape(new_shape)
+            l_vars = l_vars.reshape(new_shape)
+
+        # Getting items
+        l_means = l_means[:, key]
+        l_vars = l_vars[:, key]
 
+        # Flatten l_means and l_vars, keep the first item of the shape equal to the number of levels
         if len(l_means.shape) > 1:
```
jbrezmorf: Preform always.

src/mlmc/quantity.py
```diff
@@ -538,126 +538,113 @@ def samples(self, chunk_spec):
...
         :param key: str, int, tuple
-        :return: np.ndarray
+        :return: QuantityMean
         """
         new_qtype, start = self.qtype[key]  # New quantity type
         reshape_shape = None
-        newshape = None
+
         # ArrayType might be accessed directly regardless of qtype start and size
         if isinstance(self.qtype, qt.ArrayType):
```
jbrezmorf: getitem should be common to the Quantity and thus implemented in QType as it is the part shared between Quantity and QuantityMean. It should go as:
```
new_qtype, slice = self.qtype.get_key(key)

samples_like = data[None, None,:]     # applied for data in [self._l_means, self._l_vars]
shaped_data = self.qtype.reshape(samples_like)     # actually reshape only for Array type
new_data = shaped_data[slice].ravel()
```

src/mlmc/quantity_estimate.py
```diff
@@ -68,31 +70,16 @@ def estimate_mean(quantity, chunk_size=512000000, level_means=False):
...
-            if n > 1:
-                l_vars.append((sp - (s ** 2 / n)) / (n-1))
-            else:
-                l_vars.append((sp - (s ** 2)))
-
-    # sums full of zeros
-    if np.isnan(mean).all():
-        mean = []
-        var = []
+            l_vars.append((sp - (s ** 2)))
```
jbrezmorf: for n<=1 the variance should be `inf`.

src/mlmc/quantity_types.py
```diff
@@ -75,6 +75,9 @@ def _make_getitem_op(self, chunk, key):
         """
         return QType.keep_dims(chunk[key])
```
jbrezmorf, line 76: Considering the file name, we should possibly introduce all classes into `mlmc` module via. `mlmc.__init__.py`.

src/mlmc/quantity_types.py
```diff
@@ -136,6 +143,12 @@ def _make_getitem_op(self, chunk, key):
         chunk = chunk.reshape((*self._shape, chunk.shape[-2], chunk.shape[-1]))
         return QType.keep_dims(chunk[key])
 
+    def reshape(self, data):
+        if isinstance(self._qtype, ScalarType):
+            return data.reshape(self._shape)
```
jbrezmorf, line 148: Reshape consistently to (*self._shape, 1).

jbrezmorf: Is it compatible with reshaping the samples data?

src/mlmc/quantity.py
martinspetlik: assert

src/mlmc/quantity.py
martinspetlik: #165

src/mlmc/quantity_types.py
martinspetlik: #164

PR #160, opened 2020-12-31
Replace scalar

### LINKED ISSUE #147: replace scalar
`QType.replace_scalar`

Think it over again - try to use recursion. 
Note we should be able to replace quantity of `ScalarType` by `ArrayType`, and `ScalarType` doesn't have qtype so we need to create a new `QType` (`ArrayType`).

Suggestions from @jbrezmorf review:
`QType.replace_scalar`

- Convert to proper method of QType.
- This can be much simpler if implemented by recursive construction of the replaced type. Returning the new instance. 
- Use separate implementation for dict and scalar instead of `isinstance`.

`DictType.replace_scalar`

- How is this related to QType.replace_scalar? 
- It is not even a classmethod. It seems to be already in use, but introduction of the probably same static method in QType is confusing.
- Should be probably merged to a single method probably with virtual implementation in some QType subclases.

martinspetlik:
resolve #147

### DIFF
--- a/src/mlmc/quantity.py
+++ b/src/mlmc/quantity.py
@@ -279,7 +279,7 @@ class Quantity:
         """
         bool_type = qt.BoolType()
         new_qtype = self.qtype
-        new_qtype = qt.QType.replace_scalar(new_qtype, bool_type)
+        new_qtype = new_qtype.replace_scalar(bool_type)
         other = Quantity.wrap(other)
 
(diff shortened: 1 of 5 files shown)


martinspetlik:
resolve #147

jbrezmorf (commented):
Can be merged after MS_quantity_mean.

PR #161, opened 2021-01-09
Quantity subsampling

### LINKED ISSUE #144: Bootstrapping, subsample quantity.
Reimplement `Quantity.subsample`:
- make it operating on individual chunks without creating a list of selected samples before the chunk processing.
- avoid passing special arguments  

Target usage is [bootstrapping](https://en.wikipedia.org/wiki/Bootstrapping_(statistics)).

This makes #35 deprecated.

### LINKED ISSUE #156: Get rid of Quantity additional_params
@jbrezmorf:  We have to get rid of `additional_params`, seems that the only remaining place is the `subsample`.

### LINKED ISSUE #165: QuantityMean - Sums full of zeros leads to NaN
`QuantityMean._calculate_mean_var` contains a special case:

        if np.isnan(self._mean).all():
            self._mean = []
            self._var = []

Try to get rid of it.

martinspetlik:
resolve #156
resolve #165

### DIFF
--- a/src/mlmc/quantity.py
+++ b/src/mlmc/quantity.py
@@ -589,11 +589,6 @@ class QuantityMean:
         self._mean = np.sum(self._l_means, axis=0)
         self._var = np.sum(self._l_vars / self._n_samples[:, None], axis=0)
 
-        # Sums full of zeros leads to NaN
-        if np.isnan(self._mean).all():
-            self._mean = np.array([])
-            self._var = np.array([])
-
     @property
     def mean(self):
         if self._mean is None:
@@ -629,8 +624,6 @@ class QuantityMean:
         :return: np.ndarray, reshaped data, the final data shape depends on the particular QType
                              there is currently a reshape for ArrayType only
         """
-        if data.size == 0:
-            return data
         return self.qtype.reshape(data)
 
     def __getitem__(self, key):
--- a/src/mlmc/quantity.py
+++ b/src/mlmc/quantity.py
@@ -1,5 +1,6 @@
-import numpy as np
 import operator
+import random
+import numpy as np
 from inspect import signature
 from memoization import cached
 from typing import List
@@ -48,7 +49,6 @@ class Quantity:
         self._selection_id = self.set_selection_id()
         # Identifier of selection, should be set in select() method
         self._check_selection_ids()
-        self._op_additional_params()
 
     def get_quantity_storage(self):
         """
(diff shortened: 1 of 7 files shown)


src/mlmc/quantity.py
```diff
@@ -318,30 +303,61 @@ def ne_op(x, y):
...
+        :param chunk: np.ndarray, shape M, N, 2, where N denotes number of samples in chunk
+        :param subsample_params: instance of SubsampleParams class, it has two parameters:
+                        k: number of samples which we want to get from all chunks
+                        n: number of all samples among all chunks
         :return: np.ndarray
         """
-        quantity_storage = self.get_quantity_storage()
-        n_collected = quantity_storage.n_collected()
+        out = []
+        for _ in range(chunk.shape[1]):
```
jbrezmorf, line 318: I'm not sure that this proposed method leads to the right statistics.

src/mlmc/quantity.py
```diff
@@ -318,30 +303,61 @@ def ne_op(x, y):
...
+            i = random.randint(0, subsample_params.n)
+            if i < subsample_params.k and i < chunk.shape[1]:
+                out.append(chunk[:, i, :])
+                subsample_params.k -= 1
+            subsample_params.n -= 1
```
jbrezmorf, line 323: Clever, yet modification of the input quantity violates functional programming paradigm, which may block possible parallelization etc. If we adopt chunk iterator, then this information can be part of it. Still not the ideal solution.

src/mlmc/quantity.py
```diff
@@ -565,8 +591,8 @@ def _calculate_mean_var(self):
 
         # Sums full of zeros leads to NaN
         if np.isnan(self._mean).all():
-            self._mean = []
-            self._var = []
+            self._mean = np.array([])
```
jbrezmorf: Why

src/mlmc/quantity.py
```diff
@@ -603,6 +629,8 @@ def _reshape(self, data):
         :return: np.ndarray, reshaped data, the final data shape depends on the particular QType
                              there is currently a reshape for ArrayType only
         """
+        if data.size == 0:
```
jbrezmorf: Is it necessary?

src/mlmc/quantity_estimate.py
```diff
@@ -51,13 +52,18 @@ def estimate_mean(quantity, chunk_size=512000000):
                 # level_chunk is Numpy Array with shape [M, chunk_size, 2]
                 n_samples[level_id] += chunk.shape[1]
                 n_rm_samples[level_id] += n_mask_samples
+
+                # No samples in chunk
+                if chunk.shape[1] == 0:
```
jbrezmorf, line 61: That should not happen.

jbrezmorf (commented):
The statistical properties of the algorithm must be carefully tested or we should prove it.

src/mlmc/quantity_estimate.py
martinspetlik: It could happen. For example, if we are subsampling, a chunk might be empty.

PR #162, opened 2021-01-13
Quantity PBS run

### LINKED ISSUE #139: get rid of TransformedMomentsDerivative

### LINKED ISSUE #142: SamplingPool move dir fix
`SamplingPool.move_dir` - move condition ` if int(sample_id[-7:]) < 5: ` outside of this method. The condition is suitable just for moving several succesful samples not for moving failed samples.

### LINKED ISSUE #158: Unknow Job Id
An exception occured in `SamplingPoolPbs._qstat_pbs_job`: `Exception: qstat: Unknown Job Id 5445226.meta-pbs.metacentrum.cz`

`qstat -x *self._pbs_ids` was at blame. 
That error shouldn't ever occur, because `self._pbs_ids` contains only job ids from `qsub` run. 

I think that error occurs because there is some kind of `qstat` 'forgetfulness' of terminated jobs.

It is a very rare phenomenon, which is observed only during the long run (e.g. 8 hours for a job) of many (e.g. 2500) simulations, here it happened after around a day and a half of running MLMC.

Proposed workaround: remove job ids of long-finished jobs from `self._pbs_ids`. Perform it in periods - 24 hours should be enough.

### LINKED ISSUE #175: Quantity estimate
`mlmc.estimator.estimate` - it is unable to run `estimate_moments` and then `construct_density`. Generally, if we want to estimate moments over again, we need to recreate quantity. I find this approach inconvenient.

It doesn't work for any quantity which is the result of `quantity.subsample`.

martinspetlik:
resolve #175
resolve #158
resolve #142
resolve #139

### DIFF
--- a/src/mlmc/sample_storage_hdf.py
+++ b/src/mlmc/sample_storage_hdf.py
@@ -220,8 +220,14 @@ class SampleStorageHDF(SampleStorage):
         :return: None
         """
         for level_id, (time, n_samples) in n_ops:
+            if self._level_groups[level_id].n_ops_estimate is None:
+                self._level_groups[level_id].n_ops_estimate = [0., 0.]
+
             if n_samples > 0:
-                self._level_groups[level_id].n_ops_estimate = time/n_samples
+                n_ops_saved = self._level_groups[level_id].n_ops_estimate
+                n_ops_saved[0] += time
+                n_ops_saved[1] += n_samples
+                self._level_groups[level_id].n_ops_estimate = n_ops_saved
 
     def get_n_ops(self):
         """
@@ -230,8 +236,10 @@ class SampleStorageHDF(SampleStorage):
         """
         n_ops = list(np.zeros(len(self._level_groups)))
         for level in self._level_groups:
-            n_ops[int(level.level_id)] = level.n_ops_estimate
-
+            if level.n_ops_estimate[1] > 0:
+                n_ops[int(level.level_id)] = level.n_ops_estimate[0] / level.n_ops_estimate[1]
+            else:
+                n_ops[int(level.level_id)] = 0
         return n_ops
 
     def get_level_ids(self):
--- a/src/mlmc/sample_storage_hdf.py
+++ b/src/mlmc/sample_storage_hdf.py
@@ -4,6 +4,8 @@ from typing import List
 from mlmc.sample_storage import SampleStorage
 from mlmc.quantity_spec import QuantitySpec, ChunkSpec
 import mlmc.tool.hdf5 as hdf
+import warnings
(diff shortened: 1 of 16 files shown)


src/mlmc/sample_storage_hdf.py
```diff
@@ -218,9 +220,7 @@ def save_n_ops(self, n_ops):
         :return: None
         """
         for level_id, (time, n_samples) in n_ops:
-            if n_samples == 0:
-                self._level_groups[level_id].n_ops_estimate = 0
-            else:
+            if n_samples > 0:
```
jbrezmorf, line 230: Sure, this more correct. But what happens if there are no samples? How is n_ops_estimate initialized?

src/mlmc/sampler.py
```diff
@@ -187,7 +187,7 @@ def _store_samples(self, successful_samples, failed_samples, n_ops):
         self.sample_storage.save_samples(successful_samples, failed_samples)
         self.sample_storage.save_n_ops(n_ops)
```
jbrezmorf, line 191: Seems to big default timeout, adding samples should be pretty fast.

jbrezmorf: The default timeout should be defined as a class constant.

src/mlmc/sampling_pool.py
```diff
@@ -227,6 +228,9 @@ def _process_result(self, sample_id, result, err_msg, running_time, level_sim):
         if not err_msg:
             self._queues.setdefault(level_sim._level_id, queue.Queue()).put((sample_id, (result[0], result[1])))
             if not self._debug:
+                if int(sample_id[-7:]) < SamplingPool.N_SUCCESSFUL:
```
jbrezmorf: Make the number of saved samples a class constant, e.g. SamplingPool.n_debug_samples.

src/mlmc/tool/pbs_job.py
```diff
@@ -173,9 +173,11 @@ def calculate_samples(self):
             if not err_msg:
                 success.append((current_level, sample_id, (res[0], res[1])))
                 # Increment number of successful samples for measured time
-                SamplingPool.move_dir(sample_id, level_sim.need_sample_workspace, self._output_dir,
-                                      dest_dir=successful_dest_dir)
                 if not self._debug:
+                    # Store few successful samples
+                    if int(sample_id[-7:]) < SamplingPool.N_SUCCESSFUL:
```
jbrezmorf: Duplicit code?

src/mlmc/sampling_pool.py
martinspetlik: It is already class constant `SamplingPool.N_SUCCESSFUL`. `sample_id[-7:] `gets digits representing sample id.

PR #168, opened 2021-02-12
Chunk iterations

### LINKED ISSUE #152: Chunk iteration protocol
Chunk iteration can fully use iteration protocol, yielding the chunk (with level and possibly chunk id and storae specification along the numpy data table), the only parameter of the `samples` then would be `n_samples` possibly given indirectly through the chunk_size.

martinspetlik:
resolve #152

### DIFF
--- a/src/mlmc/quantity.py
+++ b/src/mlmc/quantity.py
@@ -112,25 +112,37 @@ class Quantity:
         """
         return self.qtype.size()
 
-    def get_cache_key(self, chunk_spec):
+    def get_cache_key(self, level_id, chunk_size=None, n_samples=None):
         """
         Create cache key
         :param chunk_spec: ChunkSpec instance
         :return: tuple
         """
-        return (chunk_spec.level_id, chunk_spec.chunk_id, chunk_spec.n_samples, chunk_spec.chunk_size, id(self))  # redundant parentheses needed due to py36, py37
+        return (level_id, chunk_size, n_samples, id(self))  # redundant parentheses needed due to py36, py37
 
-    @cached(custom_key_maker=get_cache_key)
-    def samples(self, chunk_spec):
+    #@cached(custom_key_maker=get_cache_key)
+    def samples(self, level_id=None, chunk_size=None, n_samples=None):
         """
         Yields list of sample chunks for individual levels.
         Possibly calls underlying quantities.
-        :param chunk_spec: ChunkSpec instance, it contains level_id, chunk_id and
-                                               it may contain n_samples - number of which we want to retrieve
         :return: np.ndarray or None
         """
-        chunks_quantity_level = [q.samples(chunk_spec) for q in self._input_quantities]
-        return self._operation(*chunks_quantity_level)
+        try:
+
+
+            chunks_quantity_level = [next(q.samples(level_id, chunk_size, n_samples)).data for q in self._input_quantities]
+
+            #
+            # print("len(chunk_quantity_level) ", len(chunks_quantity_level))
+            # print("chunks quantity level[0].shape ", chunks_quantity_level[0].shape)
+            #
+            # print("self._operation ", self._operation)
+
+            yield ChunkSpec(level_id=level_id, data=self._operation(*chunks_quantity_level))
+        except StopIteration:
+
+            pass
+        #return self._operation(*chunks_quantity_level)
 
     def _reduction_op(self, quantities, operation):
         """
@@ -314,6 +326,9 @@ class Quantity:
                         n: number of all samples among all chunks
         :return: np.ndarray
         """
+        print("subsample params total n ", subsample_params.total_n)
+        print("pick samples chunk shape ", chunk.shape)
+
         out = []
         for _ in range(chunk.shape[1]):
             i = random.randint(0, subsample_params.n)
@@ -341,6 +356,7 @@ class Quantity:
                 """
                 self.k = num_subsample
                 self.n = num_collected
+                self.total_n = num_collected
 
         # SubsampleParams for each level
         subsample_level_params = {key: SubsampleParams(sample_vec[key], value)
@@ -348,12 +364,12 @@ class Quantity:
         # Create a QuantityConst of dictionary in the sense of hashing dictionary items
         quantity_subsample_params = Quantity.wrap(hash(frozenset(subsample_level_params.items())))
 
-        def adjust_value(values, chunk_spec):
+        def adjust_value(values, level_id):
             """
             Custom implementation of QuantityConst.adjust_value()
             It allows us to get different parameters for different levels
             """
-            return subsample_level_params[chunk_spec.level_id]
+            return subsample_level_params[level_id]
         quantity_subsample_params._adjust_value = adjust_value
 
         return Quantity(quantity_type=self.qtype.replace_scalar(qt.BoolType()),
@@ -424,12 +440,14 @@ class Quantity:
         :return: Quantity
         """
         def _ufunc_call(*input_quantities_chunks):
+            print("input quantittes chunks ", input_quantities_chunks)
             return getattr(ufunc, method)(*input_quantities_chunks, **kwargs)
 
         quantities = []
         for arg in args:
             quantities.append(Quantity.wrap(arg))
 
+        print("quantities ", quantities)
         result_qtype = Quantity._result_qtype(_ufunc_call, quantities)
         return Quantity(quantity_type=result_qtype, input_quantities=list(quantities), operation=_ufunc_call)
 
@@ -462,7 +480,10 @@ class Quantity:
         :param method: ufunc function
         :return: QType
         """
-        chunks_quantity_level = [q.samples(ChunkSpec(level_id=0, chunk_id=0, n_samples=10)) for q in quantities]
+        try:
+            chunks_quantity_level = [next(q.samples(chunk_size=None, level_id=0, n_samples=10)).data for q in quantities]
+        except StopIteration:
+            pass
         result = method(*chunks_quantity_level)  # numpy array of [M, <=10, 2]
         qtype = qt.ArrayType(shape=result.shape[0], qtype=Quantity._get_base_qtype(quantities))
         return qtype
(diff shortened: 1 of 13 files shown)


src/mlmc/quantity_estimate.py
```diff
@@ -20,62 +19,50 @@ def cache_clear():
...
-                level_chunks_none[level_id] = True
+
+    # initialization
+    quantity_storage = quantity.get_quantity_storage()
+    level_ids = quantity_storage.level_ids()
+    n_levels = np.max(level_ids) + 1
+    n_samples = [0] * n_levels
+    n_rm_samples = [0] * n_levels
+
+    for chunk_id, chunk_slice, level_id in quantity_storage.chunks():
```
jbrezmorf: It would be simpler to pass chunk_id, chunk_slice, lavel_id as a single data class `Chunk`.

src/mlmc/quantity.py
```diff
@@ -112,31 +112,34 @@ def size(self) -> int:
         """
         return self.qtype.size()
 
-    def get_cache_key(self, chunk_spec):
+    def get_cache_key(self, level_id=None, chunk_id=None, chunk_slice=None, n_samples=None):
```
jbrezmorf: Is `n_samples` necessary? The samples chunk is uniquely given by previous parameters (proposed as a single data class Chunk.

src/mlmc/quantity.py
```diff
@@ -112,31 +112,34 @@ def size(self) -> int:
...
         """
         Create cache key
-        :param chunk_spec: ChunkSpec instance
-        :return: tuple
         """
-        return (chunk_spec.level_id, chunk_spec.chunk_id, chunk_spec.n_samples, chunk_spec.chunk_size, id(self))  # redundant parentheses needed due to py36, py37
+        chunk_size = None
+        if chunk_slice is not None:
```
jbrezmorf: Why can be `chunk_slice` None?

src/mlmc/quantity.py
```diff
@@ -112,31 +112,34 @@ def size(self) -> int:
...
         """
-        Yields list of sample chunks for individual levels.
+        Return list of sample chunks for individual levels.
         Possibly calls underlying quantities.
-        :param chunk_spec: ChunkSpec instance, it contains level_id, chunk_id and
-                                               it may contain n_samples - number of which we want to retrieve
+        :param level_id: int, level identifier
+        :param chunk_id: int, chunk identifier
+        :param chunk_slice: slice() object
+        :param n_samples: int, number of samples to retrieve
```
jbrezmorf: `n_samples` should rather be parameter of Storage.chunks

src/mlmc/quantity.py
```diff
@@ -462,7 +463,7 @@ def _result_qtype(method, quantities):
         :param method: ufunc function
         :return: QType
         """
-        chunks_quantity_level = [q.samples(ChunkSpec(level_id=0, chunk_id=0, n_samples=10)) for q in quantities]
+        chunks_quantity_level = [q.samples(level_id=0, chunk_id=0, n_samples=10) for q in quantities]
```
jbrezmorf: We should use Storage.chunks(..) and single chunk iteration here.

src/mlmc/quantity.py
```diff
@@ -527,13 +528,14 @@ def _process_value(self, value):
             value = np.array([value])
         return value[:, np.newaxis, np.newaxis]
 
-    def get_cache_key(self, chunk_spec):
+    def get_cache_key(self, level_id=None, chunk_id=None, chunk_slice=None, n_samples=None):
```
jbrezmorf: Duplicit implementation?

src/mlmc/quantity.py
```diff
@@ -677,15 +682,19 @@ def selection_id(self):
     def get_quantity_storage(self):
         return self
 
-    def samples(self, chunk_spec):
+    def chunks(self, level_id=None, n_samples=None):
```
jbrezmorf: Do we need this? Isn't Storage.chunks enough?

src/mlmc/sample_storage.py
```diff
@@ -223,30 +224,50 @@ def sample_pairs(self):
...
 
         for level_id in self.get_level_ids():
-            results = self.sample_pairs_level(ChunkSpec(level_id))
+            results = self.sample_pairs_level(level_id=level_id)
             levels_results[level_id] = results
 
         return levels_results
 
-    def sample_pairs_level(self, chunk_spec):
+    def chunks(self, level_id=None, n_samples=None):
```
jbrezmorf, line 232: Want to review storage implementation yet. But don't wait with previous comments.

src/mlmc/quantity.py
martinspetlik: `Quantity._result_qtype` calls samples method directly:

`chunks_quantity_level = [q.samples(level_id=0, chunk_id=0, n_samples=10) for q in quantities]`

There is no prior information from storage.

src/mlmc/quantity_estimate.py
martinspetlik: It's simpler, but not so readable. That's why I stop using the ChunkSpec object.

src/mlmc/quantity.py
martinspetlik: We want to get several samples (`n_samples`) to determine the results data type.

`chunks_quantity_level = [q.samples(level_id=0, chunk_id=0, n_samples=10) for q in quantities]`

src/mlmc/quantity.py
martinspetlik: I find it too complicated for such a simple use of `n_samples`.

src/mlmc/quantity.py
martinspetlik: Ok. On the other hand, I would like to keep the ability to load data directly without `Storage.chunks`

For example, to obtain data for a histogram:
`self._quantity.samples(level_id=level_id, n_samples=self._sample_storage.get_n_collected()[level_id])`

src/mlmc/quantity.py
martinspetlik: I'll try to get rid of it.

src/mlmc/quantity.py
martinspetlik: It is just a wrapper.

src/mlmc/sample_storage.py
```diff
@@ -223,30 +224,50 @@ def sample_pairs(self):
         levels_results = list(np.empty(len(np.max(self._results.keys()))))
```
jbrezmorf: The level_ids should be in fact level indices forming a sequence starting from 0 so this seems to be a bit too complicated. May the self._results be simply a list?

We should not support retrieving all samples at once as that may hang the computation when it doesn't fit into memory. Alternatively, we may support a shortcut of setting the chunk size to N followed by `samples(n_samples=N)`.

src/mlmc/sample_storage.py
```diff
@@ -223,30 +224,50 @@ def sample_pairs(self):
...
         return levels_results
 
-    def sample_pairs_level(self, chunk_spec):
+    def chunks(self, level_id=None, n_samples=None):
+        """
+        Create chunks generator
+        :param level_id: int, if not None return chunks for a given level
+        :return: generator
+        """
+        if level_id is not None:
```
jbrezmorf: Can be moved to the StorageBase.

src/mlmc/tool/hdf5.py
```diff
@@ -352,42 +349,37 @@ def scheduled(self):
...
-    def get_chunks_info(self, chunk_spec):
-        """
-        The start and end index of a chunk from a whole dataset point of view
-        :param chunk_spec: ChunkSpec instance
-        :return: List[int, int]
-        """
-        return self._chunks_info[chunk_spec]
+            if n_samples is None and chunk_slice is not None:
+                return dataset[chunk_slice]
+            elif n_samples is not None and n_samples < dataset.size:
```
jbrezmorf: Why passing the n_samples, that can be expressed by the chunk_slice.

src/mlmc/tool/hdf5.py
```diff
@@ -352,42 +349,37 @@ def scheduled(self):
...
+                return dataset[:n_samples]
+            else:
+                return dataset[()]
```
jbrezmorf: do you mean dataset[:dataset.size]?
But if the slicing handles over bound slices well returning dataset[chunk_slice] should be fine.

src/mlmc/sample_storage.py
```diff
@@ -223,30 +224,50 @@ def sample_pairs(self):
...
+            return self._results[int(level_id)].chunks()
+        return itertools.chain(*[self.level_chunks(level_id, n_samples) for level_id in self.get_level_ids()])  # concatenate generators
+
+    def level_chunks(self, level_id, n_samples=None):
```
jbrezmorf: Get rid of n_samples.

src/mlmc/quantity.py
jbrezmorf: No, it is fairly simple and significantly simplifies code on many places.

martinspetlik: ok

src/mlmc/sample_storage.py
martinspetlik: I would create `@abstractmethod` in the parent class

src/mlmc/tool/hdf5.py
martinspetlik: removed

src/mlmc/tool/hdf5.py
martinspetlik: I would like to be able to get data without calling the `chunks()` method. And this `else` is the only thing needed to do that.

src/mlmc/sample_storage.py
martinspetlik: In general, we can obtain results from a certain level before results from for example zero level. So I think this implementation is reasonable. 

Retrieving all samples at once by default is used in `Memory()` storage. I would prefer the following approach: if the user thinks that some memory problems may occur, he should use another storage.

PR #169, opened 2021-02-13
subsample random choice

### LINKED ISSUE #144: Bootstrapping, subsample quantity.

martinspetlik:
resolve #144

### DIFF
--- a/src/mlmc/quantity.py
+++ b/src/mlmc/quantity.py
@@ -1,12 +1,13 @@
 import operator
-import random
 import numpy as np
-from inspect import signature
+import scipy.stats
+import mlmc.quantity_types as qt
 from memoization import cached
 from typing import List
 from mlmc.sample_storage import SampleStorage
 from mlmc.quantity_spec import QuantitySpec
-import mlmc.quantity_types as qt
+
+RNG = np.random.default_rng()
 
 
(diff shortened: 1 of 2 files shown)


src/mlmc/quantity.py
```diff
@@ -309,24 +310,19 @@ def ne_op(x, y):
...
-        if not out:
-            out = np.array(out)[:, np.newaxis, np.newaxis]
-        return np.array(out).transpose((1, 0, 2))  # M, N, 2
+        size = scipy.stats.hypergeom(subsample_params.n, subsample_params.k, chunk.shape[1]).rvs(size=1)
+        out = RNG.choice(chunk, size=size, axis=1)
+        subsample_params.k -= out.shape[1]
+        subsample_params.n -= chunk.shape[1]
+        return out
 
     def subsample(self, sample_vec):
```
jbrezmorf, line 326: Implementation via artificial QuantityConst avoids passing Chunk spec to the operation, but it uses nasty side effects in pick_samples. In order to simplify the implementation what about a specific QuantitySubsample with the overridden samples method.

PR #195, opened 2023-01-30
Ms endorse

jbrezmorf:
Various changes made during application to the Endorse software.


mlmc/sample_storage.py
```diff
@@ -171,7 +171,7 @@ def _save_successful(self, samples):
             res = np.array(res)
             fine_coarse_res = res[:, 1]
 
-            result_type = np.dtype((np.float, np.array(fine_coarse_res[0]).shape))
```
jbrezmorf: Reason: float seems to be incompatible with new Numpy versions

mlmc/sample_storage_hdf.py
```diff
@@ -33,36 +33,14 @@ def __init__(self, file_path):
                 for i_level in range(len(self._hdf_object.level_parameters)):
                     self._level_groups.append(self._hdf_object.add_level_group(str(i_level)))
 
-    def _hdf_result_format(self, locations, times):
```
jbrezmorf, line 37: Simplified, moved into hdf5.py as it is implementation detail how the format is stored.

mlmc/sample_storage_hdf.py
```diff
@@ -96,20 +74,21 @@ def save_result_format(self, result_format: List[QuantitySpec], res_dtype):
...
+    @staticmethod
+    def make_qspec(name, unit, shape, times, locations):
+        return QuantitySpec(name.decode(), unit.decode(), shape, times, [loc.decode() for loc in locations])
 
     def load_result_format(self) -> List[QuantitySpec]:
         """
         Load result format
         """
         results_format = self._hdf_object.load_result_format()
-        quantities = []
```
jbrezmorf, line 118: Same thing, but using constructing function with named parameters.

mlmc/sample_storage_hdf.py
jbrezmorf: Moereover the format saving was wrong since it same format has been assumed for all dict elements.

mlmc/sample_storage_hdf.py
```diff
@@ -125,7 +104,8 @@ def save_samples(self, successful, failed):
     def _save_succesful(self, successful_samples):
         for level, samples in successful_samples.items():
             if len(samples) > 0:
-                self._level_groups[level].append_successful(np.array(samples))
+                ids, sample_values = zip(*samples)
```
jbrezmorf, line 125: need to make pairs here and convert to numpy arrays just before storage.

mlmc/sampling_pool.py
```diff
@@ -122,6 +122,7 @@ def calculate_sample(sample_id, level_sim, work_dir=None, seed=None):
         except Exception:
             str_list = traceback.format_exception(*sys.exc_info())
             err_msg = "".join(str_list)
+            print(err_msg)
```
jbrezmorf: TODO: replace by suitable logging? Only in case of debugging turned on. 
Problem is that the exceptions may not be propagated to the Sampler if pool somehow fail. So it would be great to log the exception within the sampling job process.

mlmc/sampling_pool.py
```diff
@@ -135,6 +136,24 @@ def change_to_sample_directory(work_dir, path: str):
         sample_dir = os.path.join(work_dir, path)
         if not os.path.isdir(sample_dir):
             os.makedirs(sample_dir, mode=0o775, exist_ok=True)
+
+        # We have observed possible problems with directory creation on the
```
jbrezmorf, line 161: I guessed that there is some problem with NFS, however in fact there was probably problem with various threads writing to the same directory, randomly cleaning it after it was created. 
I suggest to keep the code commented, it allows to better check that we are able to write into the directory. There could be problem with rights anyway.

mlmc/sampling_pool.py
```diff
@@ -241,7 +260,8 @@ def _process_result(self, sample_id, result, err_msg, running_time, level_sim):
         self._save_running_time(level_sim._level_id, running_time)
 
         if not err_msg:
-            self._queues.setdefault(level_sim._level_id, queue.Queue()).put((sample_id, (result[0], result[1])))
```
jbrezmorf: just to improve readibility

mlmc/sampling_pool_pbs.py
```diff
@@ -200,49 +212,49 @@ def execute(self):
         Execute pbs script
         :return: None
         """
-        if len(self._scheduled) > 0:
```
jbrezmorf, line 213: Decrease indention level

mlmc/sampling_pool_pbs.py
```diff
@@ -200,49 +212,49 @@ def execute(self):
...
-            script_content = "\n".join(self.pbs_script)
-            self.write_script(script_content, job_file)
+            # Get pbs_id from qsub output
+            pbs_id = process.stdout.split(".")[0]
+            # Store pbs id for future qstat calls
+            self._pbs_ids.append(pbs_id)
+            pbs_process.write_pbs_id(pbs_id)
 
-            process = subprocess.run(['qsub', job_file], stderr=subprocess.PIPE, stdout=subprocess.PIPE)
-            try:
```
jbrezmorf: I quite didn't get reason for this try block.  What is reason for PIPEs? I propose to just let subproces.run finish and take resulting stdout and stderr.

mlmc/tool/hdf5.py
```diff
@@ -1,6 +1,30 @@
 import numpy as np
 import h5py
 from mlmc.quantity.quantity_spec import ChunkSpec
+import time
+import logging
+
+class FileSafe(h5py.File):
```
jbrezmorf, line 7: Context manager to open the file with some timeout.

mlmc/sample_storage_hdf.py
```diff
@@ -33,36 +33,14 @@ def __init__(self, file_path):
                 for i_level in range(len(self._hdf_object.level_parameters)):
```
jbrezmorf, line 37: Changes in StorageHDF need to be made in StorageMemory as well.

mlmc/tool/hdf5.py
```diff
@@ -139,7 +168,24 @@ def result_format_dset_name(self):
         """
         return "result_format"
 
-    def save_result_format(self, result_format, res_dtype):
+    def single_format(self, spec:"QuantitySpec"):
```
jbrezmorf: Moved from sample_storage_hdf5 as QunatitySpec serialization.

mlmc/tool/hdf5.py
```diff
@@ -150,33 +196,26 @@ def save_result_format(self, result_format, res_dtype):
         :param res_dtype: numpy.dtype used for the dataset storage of a single QuantitySpec
         :return: None
         """
-        result_format_dtype = res_dtype
-
-        # Ensure dataset exists (resizable)
         with h5py.File(self.file_name, 'a') as hdf_file:
+            # format item in main group
```
jbrezmorf: Switch to hdf5 group in order to support heterogeneous items in the result.

mlmc/tool/simple_distribution.py
```diff
@@ -144,7 +144,10 @@ def density(self, value):
         moms = self.eval_moments(value)
         power = -np.sum(moms * self.multipliers / self._moment_errs, axis=1)
         power = np.minimum(np.maximum(power, -200), 200)
-        return np.exp(power)
```
jbrezmorf, line 147: Fix of error caused by more strict new Numpy versions that disallow automatic conversion of (1,)shape to scalar.

mlmc/sampling_pool.py
jbrezmorf: We do not have support for worker job logging. So keep it.

mlmc/sampling_pool_pbs.py
```diff
@@ -210,52 +223,71 @@ def execute(self):
...
+        # mechanism using self._qsub_failed_n and the outer polling loop in
+        # mlmc.sampler.Sampler.ask_sampling_pool_for_samples.
+        process = self.pbs_commands.qsub([job_file])
+        # AGENT: following try block is present in master and base but
+        # was removed in this branch, is it sufficiently replaced in pbs_commands.qsub ?
+        # Resolved: PbsCommands returns a decoded CommandOutput for completed
+        # commands; qsub failures are handled by returncode below, while command
+        # execution errors still propagate as environment errors.
+        # try:
+        if process.returncode != 0:
```
jbrezmorf, line 252: Try block removed.
 PbsCommands returns a decoded CommandOutput for completed
 commands; qsub failures are handled by returncode below, while the command
 execution errors still propagate as environment errors.

